PR #33, opened 2023-12-19
chore: package/cross-compile

markphelps:
getting cross compiliation/packaging of engine for multiple architectures

currently, this builds and zips up the engine for the following os/architectures:
- linux/x86_64
- linux/arm64
- mac/arm64

The plan is to integrate this with #35 or similar so that the process is

- use https://github.com/softprops/action-gh-release or release please to create release
- this workflow will then build the lib for the various architectures and upload them to the new release
- we'll then have another workflow download these uploaded zip files and package each of the clients in preparation for publishing

### DIFF
--- a/.github/workflows/build-engine.yml
+++ b/.github/workflows/build-engine.yml
@@ -29,6 +29,16 @@ jobs:
           target: ${{ matrix.platform.target }}
           override: true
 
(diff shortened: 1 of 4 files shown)


GeorgeMac (approved):
Delightful

Issue #38, opened 2023-12-20
Package + Distribute


markphelps:
Figure out how to build and publish native packages instead of requiring users to build from source

## TODO

- [x] Figure out how to package per ecosystem
  - [x] Ruby
  - [x] Go
  - [x] Node
  - [x] Python
- [x] Publish to each ecosystem (gitea packages for testing)
  - [x] Ruby
  - [x] Go
  - [x] Node
  - [x] Python
- [x] Publish to real package registries for each ecosystem
  - [x] Ruby
  - [x] Go
  - [x] Node
  - [x] Python
- [x] Test
- [x] Update Docs

markphelps:
Done

PR #39, opened 2023-12-20
Build/Package

markphelps:
Re: #38 

First steps in packaging natively per ecosystem

This creates a dagger pipeline to build packages for each ecosystem.

It currently assumes that the built libraries for each arch/platform exist locally in the following structure

```
tmp
├── darwin_arm64
│   └── libfliptengine.dylib
├── flipt_engine.h
├── linux_arm64
│   └── libfliptengine.so
└── linux_x84_64
    └── libfliptengine.so
```

This will be adjusted once we update this pipeline to run in GH Actions and in combo with #33 

This also updates the tests to depend on these files being in the above mentioned structure and drops windows support for now

## TODO

- [x] Figure out what to do with the Makefile now that the output paths have changed for the lib
- [x] Update per client readmes

### DIFF
--- a/build/main.go
+++ b/build/main.go
@@ -5,7 +5,10 @@ import (
 	"os"
 
 	"dagger.io/dagger"
+	"golang.org/x/sync/errgroup"
 )
+
+var sema = make(chan struct{}, 5)
 
 func main() {
 	err := run()
@@ -13,6 +16,8 @@ func main() {
 		panic(err)
 	}
 }
+
+type buildFn func(context.Context, *dagger.Client, *dagger.Directory) error
 
 func run() error {
 	ctx := context.Background()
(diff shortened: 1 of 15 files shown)


build/main.go
```diff
@@ -0,0 +1,113 @@
...
+func main() {
+	err := run()
+	if err != nil {
+		panic(err)
+	}
+}
+
+type buildFn func(context.Context, *dagger.Client, *dagger.Directory) error
+
+func run() error {
```
yquansah, line 22: Should we add a flag argument to select languages you wish to build as well?

markphelps: yeah can do that.

GeorgeMac (approved):
Looks good

build/main.go
markphelps: will do in a follow up

PR #79, opened 2024-01-03
refactor(engine): use evaluation.proto as source of truth for deseria…

erka:
…lize Flipt response


yquansah (commented):
This is amazing work @erka 👏! Having one source of truth for the data models seems appropriate.

flipt-engine/proto/evaluation.proto
```diff
@@ -0,0 +1,200 @@
...
+}
+
+message EvaluationNamespaceSnapshotRequest {
+  string key = 1;
+}
+
+// flipt:sdk:ignore
+service DataService {
+  rpc EvaluationSnapshotNamespace(EvaluationNamespaceSnapshotRequest) returns (EvaluationNamespaceSnapshot);
+}
```
yquansah, line 200: This file could potentially be a moving target if things change in the original repository, just to point it out.

flipt-engine/src/models/common.rs
```diff
@@ -9,6 +9,15 @@ pub enum FlagType {
     Boolean,
 }
 
+impl From<i32> for FlagType {
+    fn from(n: i32) -> FlagType {
+        match n {
+            1 => FlagType::Boolean,
+            _ => FlagType::Variant,
```
yquansah, line 16:
```suggestion
            0 => FlagType::Variant,
            _ => FlagType::Variant,
```

Maybe we should be all the way explicit here.

flipt-engine/src/models/common.rs
```diff
@@ -18,6 +27,15 @@ pub enum SegmentOperator {
     And,
 }
 
+impl From<i32> for SegmentOperator {
+    fn from(n: i32) -> SegmentOperator {
+        match n {
+            1 => SegmentOperator::And,
+            _ => SegmentOperator::Or,
```
yquansah, line 34:
```suggestion
            0 => SegmentOperator::Or,
            _ => SegmentOperator::Or,
```

Same here, maybe we should be all the way explicit.

flipt-engine/src/models/common.rs
```diff
@@ -26,6 +44,14 @@ pub enum SegmentMatchType {
     #[serde(rename = "ALL_SEGMENT_MATCH_TYPE")]
     All,
 }
+impl From<i32> for SegmentMatchType {
+    fn from(n: i32) -> SegmentMatchType {
+        match n {
+            1 => SegmentMatchType::Any,
+            _ => SegmentMatchType::All,
```
yquansah, line 51:
```suggestion
            0 => SegmentMatchType::All,
            _ => SegmentMatchType::All,
```

Same here.

flipt-engine/proto/evaluation.proto
erka: Yep, ideally it would be to have this in `flipt-rust` client and use it in this project. What do you think about this @markphelps?

markphelps: yeah, we're actively refactoring our Server side SDKs at the moment (rust included), so I would hold off for now

Also the `DataService` is really meant to be 'internal' in that users shouldn't depend on its structure (although we will always aim for backward compatibility in reality because protos).

tl;dr I think this is fine for now

flipt-engine/examples/grps.rs
```diff
@@ -0,0 +1,29 @@
+// cargo run --example grps
```
markphelps, line 1: should this file be named `grpc`?

flipt-engine/gen/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "protos"
```
markphelps, line 2: this doesnt get built into the `engine` lib correct? just want to ensure the engine libraries are as small as they can be since they are embedded in all the clients

markphelps (commented):
great idea @erka !!! just a couple minor questions

flipt-engine/examples/grps.rs
erka: You're right :)

flipt-engine/gen/Cargo.toml
erka: With this goal we don't need `tonic`. I built release for main and this branches. It's results.
```
// main
768B libfliptengine.d
8.2M libfliptengine.dylib
1.5M libfliptengine.rlib

// this 
928B libfliptengine.d
8.3M libfliptengine.dylib
1.3M libfliptengine.rlib
```

PR #91, opened 2024-01-11
refactor: move evaluation into own library crate

GeorgeMac:
I don't know if this is a good idea or not, let alone if it is going to work.

This PR moves the core evaluation logic into a library sub-crate.

The purpose is to ultimately be able to create a WASM target alongside the C dylib target.
For now, I have left the root of the project as the dylib target.
My thought is, next another subcrate which depend on the `flipt_evaluation` crate can be created.
This can be the one that exposes functions with wasm-bindgen.

This was also an exercise in me getting used to Rust 😂 
All feedback / pushback welcomed.


markphelps (approved):
happy crabbin!

Issue #95, opened 2024-01-15
[FLI-777] Preserve previous snapshot if parser fails


markphelps:
https://github.com/flipt-io/flipt-client-sdks/blob/0432b9276471e4d310cc92db53931dae07b66315/flipt-engine/lib/evaluation/src/store/mod.rs#L50

Instead of blanking out the state with a new Document, we should preserve the last known state.

<sub>[FLI-777](https://linear.app/flipt/issue/FLI-777/preserve-previous-snapshot-if-parser-fails)</sub>

Issue #60, opened 2024-01-02
[FLI-769] Allow for debug logging in engine


markphelps:
We need a way to enable debug level logging for the engine, for example to see:

- what `namespaces/flags` the engine is aware of
- the results of requests to Flipt server
- the results of evaluation for variant/boolean

I think this will need to be implemented in two parts:

1. Add debug logging to the rust engine guarded
2. Add a way to configure this logging level per each client and pass that down to the rust engine

Instead of simply enabling `DEBUG` or not, we could think about how to support X number of log levels, like `DEBUG`, `INFO`, `WARN`, `ERROR` for example and standardize on those

<sub>[FLI-769](https://linear.app/flipt/issue/FLI-769/allow-for-debug-logging-in-engine)</sub>

markphelps:
done

PR #96, opened 2024-01-15
fix: dont wipe out snapshot if error in parser

markphelps:
Fixes: #95 

- Modifies the `Snapshot::build` to propagate any errors instead of returning an empty Document

We still swallow the error in `replace_snapshot`, and will need to figure out what we want to do here, likely something related to #60 

This also returns an error on first call to `new_snapshot_evaluator` at engine initialization if it can't do the first call to get the snapshot from the API, which I think is likely the behavior we want for now


yquansah (approved):
Yeah i think this is fine for now. We will figure out logging, and other error scenarios later on.

PR #116, opened 2024-01-23
feat(authentication): add support for substitutable authentication methods

yquansah:
This PR adds support for substitutable authentication methods. The purpose of this is for the client to either provide a Bearer or JWT token.

Completes FLI-763

### DIFF
--- a/flipt-client-go/README.md
+++ b/flipt-client-go/README.md
@@ -30,7 +30,7 @@ func main() {
   // flipt.WithNamespace(string): configures which namespace you will be making evaluations on
   // flipt.WithURL(string): configures which upstream Flipt data should be fetched from
   // flipt.WithUpdateInterval(int): configures how often data should be fetched from the upstream
-  // flipt.WithAuthToken(string): configures an auth token if your upstream Flipt instance requires it
+  // flipt.WithAuthentication(any): configures an authentication method if your upstream Flipt instance requires it
   evaluationClient, err := flipt.NewClient()
   if err != nil {
     log.Fatal(err)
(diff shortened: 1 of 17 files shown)


flipt-evaluation/src/parser/mod.rs
```diff
@@ -1,34 +1,71 @@
...
+
 use crate::error::Error;
 use crate::models::source;
 
 pub trait Parser {
     fn parse(&self, namespace: &str) -> Result<source::Document, Error>;
 }
 
+#[derive(Deserialize, Clone)]
+pub struct Authentication {
```
markphelps, line 12: could this be an enum since its an either or kind of thing and not both?

flipt-client-go/README.md
```diff
@@ -30,7 +30,7 @@ func main() {
   // flipt.WithNamespace(string): configures which namespace you will be making evaluations on
   // flipt.WithURL(string): configures which upstream Flipt data should be fetched from
   // flipt.WithUpdateInterval(int): configures how often data should be fetched from the upstream
-  // flipt.WithAuthToken(string): configures an auth token if your upstream Flipt instance requires it
+  // flipt.WithAuthentication(any): configures an authentication method if your upstream Flipt instance requires it
```
markphelps, line 33: just FYI this is gonna be a breaking change across all clients, which I think is ok since we are in the `0.0.x` range, but this will likely be the `0.1.0` release

yquansah: @markphelps Yeah noted!

flipt-evaluation/src/parser/mod.rs
yquansah: @markphelps Good suggestion. I made the change to use an `enum` instead. Thank you!

markphelps (approved):
Looks good ! I can do the release in the morning

PR #118, opened 2024-01-24
chore(authentication): fix camelcasing of jwt token

yquansah:
Fix the camel casing of jwt token, and add tests for correctness of enum.


markphelps (approved):
Good catch

PR #128, opened 2024-02-06
feat(client-sdks): add batch evaluation to engine

yquansah:
Add batch evaluation to engine + clients.

This was missing previously, so we are adding it for users to be able to leverage the batch evaluation functionality.

### DIFF
--- a/flipt-client-go/models.go
+++ b/flipt-client-go/models.go
@@ -7,6 +7,12 @@ type evaluationRequest struct {
 	Context      map[string]string `json:"context"`
 }
 
+type InputEvaluationRequest struct {
+	FlagKey  string            `json:"flag_key"`
+	EntityId string            `json:"entity_id"`
+	Context  map[string]string `json:"context"`
+}
+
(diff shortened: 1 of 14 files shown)


yquansah:
Just got to do ruby. I'll attack that tomorrow lol

flipt-client-go/models.go
```diff
@@ -7,6 +7,12 @@ type evaluationRequest struct {
 	Context      map[string]string `json:"context"`
 }
 
+type InputEvaluationRequest struct {
```
markphelps, line 10:
```suggestion
type EvaluationRequest struct {
```

markphelps: wdyt about just EvaluationRequest for all these?

yquansah: @markphelps Yeah I didnt want to expose the fact that you can input a namespace. But there is a cleaner way to do this. Pushed up another commit.

markphelps (approved):
one minor change, otherwise lgtm!! thanks

Issue #81, opened 2024-01-03
[FLI-771] Java Client Side SDK


markphelps:
### Language

Java

### FFI Support

- [X] I've validated that the language supports Foreign Function Interface (FFI) bindings

### Additional Context

https://developer.okta.com/blog/2022/04/08/state-of-ffi-java

https://blog.payara.fish/java-21-foreign-function-and-memory-api

### Willing to Implement

- [X] I'm willing to help implement the Flipt Client SDK in this language if it's accepted. 🙌

### Willing to Review

- [X] I'm willing to review contributions to the Flipt Client SDK in this language if it's accepted.

<sub>[FLI-771](https://linear.app/flipt/issue/FLI-771/java-client-side-sdk)</sub>

PR #134, opened 2024-02-13
feat(java): Add java SDK for client sdks

yquansah:
Adding Java client SDK. Closes: https://github.com/flipt-io/flipt-client-sdks/issues/81

Completes FLI-771


flipt-client-java/.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
+/shelf/
```
markphelps, line 2: can we exclude all `.idea/` since its editor specific?

flipt-client-java/build.gradle
```diff
@@ -0,0 +1,86 @@
+plugins {
+    id 'java-library'
+    id 'maven-publish'
+    id "com.diffplug.spotless" version "6.25.0"
+}
+
+group = 'io.flipt'
+version = '1.0.0'
```
markphelps, line 8:
```suggestion
version = '0.1.0'
```

flipt-client-java/src/main/java/io/flipt/client/models/EngineOpts.java
```diff
@@ -0,0 +1,44 @@
...
+    this.reference = reference;
+  }
+
+  @JsonProperty("url")
+  public Optional<String> getUrl() {
+    return url;
+  }
+
+  @JsonProperty("update_interval")
+  public Optional<Integer> getUpdateInterval() {
```
markphelps, line 31: Can we make this a `Duration` ? https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html

flipt-client-java/src/main/java/io/flipt/client/models/EvalRequest.java
```diff
@@ -0,0 +1,42 @@
+package io.flipt.client.models;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Map;
+
+public class EvalRequest {
```
markphelps, line 6: can we make this private or protected? I don't think we want to expose this in the public API

flipt-client-java/build.gradle
```diff
@@ -0,0 +1,86 @@
...
+
+    if (osName.contains("linux") || osName.contains("nix")) {
+        String archName = System.getProperty("os.arch").toLowerCase()
+        if (archName == "amd64" || archName == "x86_64") {
+            systemProperty("jna.library.path", System.getProperty("user.dir")+File.separator+"lib"+File.separator+"ext"+File.separator+"linux_x86_64")
+        }
+        else if (archName == "aarch64" || archName == "arm64") {
+            systemProperty("jna.library.path", System.getProperty("user.dir")+File.separator+"lib"+File.separator+"ext"+File.separator+"linux_arm64")
+        }
+    }
```
markphelps, line 85: is it possible to make this fail the build if none of these are true?

markphelps (commented):
couple minor things. looks great!! thank you for tackling this. you are the java expert now on the team

flipt-client-java/src/main/java/io/flipt/client/models/EvalRequest.java
yquansah: @markphelps great idea!

flipt-client-java/build.gradle
yquansah: Did so! Great idea

markphelps (approved):
looks great to me!!!

PR #135, opened 2024-02-14
chore(java): add dynamic library to resources directory

yquansah:
Add shared library to resources directory, so it can be included in the jar archive.


markphelps (approved):
looks good! we can delete the existing tag and repush

PR #138, opened 2024-02-16
fix: put libs in correct path expected by JNA

markphelps:
Attempt to fix:

`Native library (linux-x86-64/libfliptengine.so) not found in resource path`

I think this happens because JNA behaves differently when loading from the local file system (ie when running tests) than it does when the lib is packaged in a JAR

See: https://github.com/java-native-access/jna/blob/master/www/GettingStarted.md

The relevant part for us is

> Make your native library available on your classpath, under the path {OS}-{ARCH}/{LIBRARY}, where {OS}-{ARCH} is JNA's canonical prefix for native libraries (e.g. win32-x86, linux-amd64, or darwin). If the resource is within a jar file it will be automatically extracted when loaded.

I found the structure it expects for each OS/Arch combo, which does not match our current directory layout in `/src/resources/`
CleanShot 2024-02-15 at 11 12 26

**Note** : this image actually shows them in the wrong place because it was an older pre-release.. but just imagine the libs are in `/src/main/resources`...

Here is the test which shows the way JNA expects the OS/Arch structures to look like: https://github.com/java-native-access/jna/blob/master/test/com/sun/jna/PlatformTest.java#L50

When we build the package in `build/main.go`, we previously downloaded the built engine libs and just put them in the `src/main/resources/` folder in the container with the above structure.

The fix is to rename this structure to one that JNA expects before mounting them in the container:

```
	// the directory structure of the tmp directory is as follows:
	// tmp/linux_x86_64/
	// tmp/linux_arm64/
	// tmp/darwin_arm64/

	// we need to convert it to the following structure:
	// tmp/linux-x86-64/
	// tmp/linux-arm/
	// tmp/darwin-aarch64/
```


yquansah (approved):
Nice one I really appreciate this.

PR #139, opened 2024-02-18
chore(breaking): java rename shutdown to close to be consistent

markphelps:
java rename shutdown to close to be consistent with other libs; add close to Java example


yquansah (approved):
Yeah I was thinking this too. Good call thanks mark

PR #145, opened 2024-02-19
feat(evaluation): add entityId for constraint comparisons

yquansah:
Add entityId for constraint comparisons.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,54 @@
+LIB = libfliptengine
+HEADER = flipt_engine.h
+
+.DEFAULT_GOAL := help
+.PHONY: build go node python ruby java help clean
+
+# get os and arch
+OS := $(shell uname -s | tr A-Z a-z)
+ARCH := $(shell uname -m | tr A-Z a-z)
+
+build: ## Build the shared engine library
+	cargo build --release
+
+go: build ## Build the go client package
+	mkdir -p flipt-client-go/ext/$(OS)_$(ARCH)
+	cp target/release/$(LIB).* flipt-client-go/ext/$(OS)_$(ARCH)/
+	cp target/release/$(HEADER) flipt-client-go/ext/
+	go install ./flipt-client-go
+
+node: build ## Build the node client package
+	mkdir -p flipt-client-node/ext/${OS}_${ARCH}
+	cp target/release/$(LIB).* flipt-client-node/ext/${OS}_${ARCH}/
+	cd flipt-client-node && npm install && npm run build && npm pack
+
+python: build ## Build the python client package
+	mkdir -p flipt-client-python/ext/${OS}_${ARCH}
+	cp target/release/$(LIB).* flipt-client-python/ext/${OS}_${ARCH}/
+	cd flipt-client-python && poetry build
+
+ruby: build ## Build the ruby client gem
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -0,0 +1,54 @@
...
+	rm -rf flipt-client-node/dist
+	rm -rf flipt-client-python/ext/*
+	rm -rf flipt-client-python/dist
+	rm -rf flipt-client-ruby/lib/ext/*
+	rm -rf flipt-client-ruby/pkg/
+	rm -rf flipt-client-java/src/main/resources/*
+
+help:
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk \
+		'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
```
yquansah, line 54: Not sure what happened but the Makefile got missing

markphelps: i deleted it because it was a pain to keep updating and didn't know anybody else found it useful. Happy to keep it though

flipt-evaluation/src/models/common.rs
```diff
@@ -40,6 +40,8 @@ pub enum ConstraintComparisonType {
     Boolean,
     #[serde(rename = "DATETIME_CONSTRAINT_COMPARISON_TYPE")]
     DateTime,
+    #[serde(rename = "ENTITY_ID_CONSTRAINT_COMPARISON_TYPE")]
+    EntityId,
```
markphelps, line 44: I think we can make `Unknown` the fallback value by moving it to the last variant of this enum and tagging it like so:

```rust
#[derive(Clone, Debug, Default, Deserialize, PartialEq)]
pub enum ConstraintComparisonType {
    #[serde(rename = "STRING_CONSTRAINT_COMPARISON_TYPE")]
    String,
    #[serde(rename = "NUMBER_CONSTRAINT_COMPARISON_TYPE")]
    Number,
    #[serde(rename = "BOOLEAN_CONSTRAINT_COMPARISON_TYPE")]
    Boolean,
    #[serde(rename = "DATETIME_CONSTRAINT_COMPARISON_TYPE")]
    DateTime,
    #[serde(rename = "ENTITY_ID_CONSTRAINT_COMPARISON_TYPE")]
    EntityId,
    #[default]
    #[serde(other)]
    #[serde(rename = "UNKNOWN_CONSTRAINT_COMPARISON_TYPE")]
    Unknown,
}
```

https://serde.rs/variant-attrs.html#other

markphelps: we should likely do this for all enum values that we can to prevent the panic errors on future additions

yquansah: @markphelps Done!

Makefile
yquansah: @markphelps ohhh i definitely use it 😂

flipt-evaluation/src/models/common.rs
markphelps: @yquansah can you add a test to show that it successfully deserializes to `Unknown` given a value not in this enum in JSON pls

markphelps (approved):
nice

Issue #137, opened 2024-02-16
[FLI-848] Add `ListFlags` support


markphelps:
From Discord:

Add `ListFlags` method, returning all available flags for evaluation

<sub>[FLI-848](https://linear.app/flipt/issue/FLI-848/add-listflags-support)</sub>

PR #151, opened 2024-02-25
feat: list flags + single namespace

markphelps:
Fixes: #137 

I also am looking into how we may be able to simplify things if we move away from passing `vec[namespaces]` everywhere and only hold a single namespace in the evaluator, as that's all we expose from the outer clients anyways

## TODO

**Note:** I only exposed implemented `list_flags` in Ruby right now until I get some feedback on the approach

- [x] Add `list_flags` to other languages/clients
  - [x] node
  - [x] python
  - [x] ruby
  - [x] go
  - [x] java
- [x] Add more tests
- [x] Figure out response type/payload for `list_flags`


flipt-client-ruby/spec/evaluation_spec.rb
```diff
@@ -90,4 +90,15 @@
...
   end
+
+  describe '#list_flags' do
+    it 'returns a list of flags' do
+      resp = @client.list_flags
+
+      expect(resp).to_not be_nil
+      expect(resp['status']).to eq('success')
+      expect(resp['error_message']).to be_nil
+      expect(resp['result']).to include({"enabled" => true, "key" => "flag_boolean", "type" => "BOOLEAN_FLAG_TYPE"})
```
markphelps, line 101: idk if I love using the same response object that we use for evaluation here.. wdyt?

markphelps:
@yquansah wdyt?

yquansah (commented):
This is great for the `list_flags`, nice implementation.

The only gripe I have with simplifying client to accept one namespace is that its not really hurting anything currently to have the code right now at least be in a place to potentially accept multiple namespaces if the time ever comes.

flipt-client-ruby/spec/evaluation_spec.rb
yquansah: @markphelps What did you have in mind to change it to?

markphelps: Not sure, just trying to think from a consumers point of view. As right how This is only used for evaluation, but if we use the same structure for list flags it might be a bit confusing? I guess we could just namespace it by doing `result.flags` which would return the array of the flags?

markphelps:
cc @piclemx

flipt-client-python/flipt_client/models.py
```diff
@@ -1,6 +1,8 @@
 from pydantic import BaseModel
 from typing import List, Optional
 
+import pydantic
```
GeorgeMac, line 4:
```suggestion
from pydantic import BaseModel, RootModel
from typing import List, Optional
```

flipt-evaluation/src/lib.rs
```diff
@@ -161,15 +162,30 @@ where
...
+            Some(f) => f,
+            None => {
+                return Err(Error::Unknown(format!(
+                    "failed to get flags for {}",
+                    self.namespace,
+                )));
+            }
+        };
+
+        Ok(flags)
```
GeorgeMac, line 177: Take it or leave it (im also guessing based on what I think I remember):
```suggestion
        match self.store.list_flags(&self.namespace) {
            Some(f) => Ok(f),
            None => Err(Error::Unknown(format!(
                "failed to get flags for {}",
                self.namespace,
            ))),
        };
```

GeorgeMac (approved):
I sprinkled in a couple take it or leave its

flipt-client-python/flipt_client/models.py
markphelps: i tried that previously and it was giving me troubles about it couldn't find that import, will try again tho

PR #154, opened 2024-02-28
go cleanup memory after use

erka:
Based on https://go.dev/blog/cgo#strings-and-things


markphelps (approved):
thanks @erka! I had wondered about this while I was working on #151 and meant to circle back to it

markphelps:
lol oh now I see this was on top of #151 . will get that into a reviewable state tonight

erka:
FFI is a challenge and there is a lot of hidden gems :(

PR #157, opened 2024-03-02
fix(engine-ffi): prevent memory leaks

erka:
todos:
- [x] create the function to clean the pointer for  CString::into_raw()
- [x] update the clients to return the const *char to ffi for memory cleanup

> [CString::into_raw](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.into_raw) The pointer which this function returns must be returned to Rust and reconstituted using [CString::from_raw](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.from_raw) to be properly deallocated. Specifically, one should not use the standard C free() function to deallocate this string.


yquansah (approved):
Ah understood, so since we are returning a "raw" string or so, the memory needs to be cleaned up on the Rust side as well.

markphelps (approved):
ffi..

thanks @erka !!

Issue #165, opened 2024-03-14
[Bug]: Local evaluation does not have same behavior as remote


SamuelCabralCruz:
### Bug Description

Two flags with same configuration have different value using local evaluation.

### Version Info

1.38.0

### Search

- [X] I searched for other open and closed issues before opening this

### Steps to Reproduce

Here is a configuration to reproduce the issue:

```yaml
# exported by Flipt (v1.38.0) on 2024-03-13T19:23:31Z
version: "1.2"
namespace: default
flags:
- key: 3410-redesign
  name: redesign
  type: BOOLEAN_FLAG_TYPE
  enabled: false
  rollouts:
  - segment:
      key: 0000-4a845c25fa825feac1813bf514e0ca76
      value: true
- key: 3410-redesign-dashboard
  name: redesign-dashboard
  type: BOOLEAN_FLAG_TYPE
  enabled: false
  rollouts:
  - segment:
      key: 0000-4a845c25fa825feac1813bf514e0ca76
      value: true
- key: 2697-internal-conversion
  name: 2697-internal-conversion
  type: BOOLEAN_FLAG_TYPE
  enabled: false
  rollouts:
  - segment:
      key: 0000-4a845c25fa825feac1813bf514e0ca76
      value: true
- key: 2709-event-v2
  name: 2709-event-v2
  type: BOOLEAN_FLAG_TYPE
  enabled: false
  rollouts:
  - segment:
      key: 0000-4a845c25fa825feac1813bf514e0ca76
      value: true
- key: 2709-sso-okta
  name: sso-okta
  type: BOOLEAN_FLAG_TYPE
  enabled: false
  rollouts:
  - segment:
      key: 0000-4a845c25fa825feac1813bf514e0ca76
      value: true
- key: 3049-orders-notes
  name: orders-notes
  type: BOOLEAN_FLAG_TYPE
  enabled: false
  rollouts:
  - segment:
      key: 0000-4a845c25fa825feac1813bf514e0ca76
      value: true
- key: 3049-search-v2
  name: search-v2
  type: BOOLEAN_FLAG_TYPE
  enabled: false
  rollouts:
  - segment:
      key: 0000-4a845c25fa825feac1813bf514e0ca76
      value: true
- key: 3049-orders-v2
  name: orders-v2
  type: BOOLEAN_FLAG_TYPE
  enabled: false
  rollouts:
  - segment:
      key: 0000-4a845c25fa825feac1813bf514e0ca76
[...]

PR #166, opened 2024-03-14
fix: remove continue for properties that are not in map

yquansah:
This will fix the issues of bad evaluation for `notpresent` and `present` operators.

Closes: https://github.com/flipt-io/flipt-client-sdks/issues/165


markphelps (approved):
ty!

PR #175, opened 2024-03-22
refactor: Prep for Wasm

markphelps:
This PR refactors the code to extract all of the FFI specific stuff into the `flipt-engine-ffi` lib and try to make `flipt-evaluation` a more general purpose library that defines some contracts (traits) and has the core evaluation logic as public functions.

The reasoning behind this is to get all the stuff that wont work in WASM (concurrency) out of the `flipt-evaluation` code so that it can (hopefully) be used by a new `flipt-engine-wasm` lib that will be compiled down to WASM. Wasm Wasm.

The `flipt-engine-ffi` lib is now responsible for:
- Implementing a thread safe evaluator
- Implementing an HTTP Parser (might rename this in the future) to query Flipt Data API for flag state

The `fliptevaluation` lib is now only responsible for:
- Defining common data structures
- Building a Snapshot from a document (JSON)
- Implementing `Store` for Snapshot
- Performing the actual evaluation given data passed into it

cc @erka


GeorgeMac (approved):
Nice

PR #178, opened 2024-03-23
fix(flipt-evaluation): increment bucket search index by one

GeorgeMac:
See: https://github.com/flipt-io/flipt/issues/2890#issuecomment-2016430657

This brings our Rust implementation inline with our Go implementation of search.


GeorgeMac:
I knew just doing this in the GH ui would be a mistake 😂

markphelps (approved):
+1

Issue #187, opened 2024-03-25
Request to Include Flipt Engine Header File for Go Client SDK


sylwang:
Hi there!

My team has been running into an issue below:
```
2024-03-25 13:43:21 # go.flipt.io/flipt-client
2024-03-25 13:43:21 /cache/go.flipt.io/flipt-client@v0.4.3/evaluation.go:9:10: fatal error: ./ext/flipt_engine.h: No such file or directory
2024-03-25 13:43:21     9 | #include "./ext/flipt_engine.h"
2024-03-25 13:43:21       |          ^~~~~~~~~~~~~~~~~~~~~~
2024-03-25 13:43:21 compilation terminated.
```

We imported the Go client SDKs and included it in our `go.mod` file for our project. Since there is a requirement for including the flipt engine header file, we had to fork the flipt client SDKs into our project and then manually modify the `/ext` directory to include this header file.

Golang doesn't allow modifications to an imported library, so the only way to solve this problem is to fork the repo and manually include the header file in `/ext` directory. That is something we want to avoid doing for our project. 

What are your suggestions? And I wonder Is there a reason why this flipt engine header file wasn't included (if it is just a static header file) for the Go client SDKs?

Thank you!

markphelps:
Thanks for reporiting @sylwang !

I may have a fix in #188 , unfortunately the only real way I can test it is by creating a new release of the Go SDK, because of how go works with modules/github/etc.

We have to effectively copy the subdirectory of `flipt-client-go` from this repo and push it to it's own repo: https://github.com/flipt-io/flipt-client-sdks/blob/a44168b902bb20092174a209df59e03ba8a7652f/build/main.go#L142-L148

https://github.com/flipt-io/flipt-client-go

Im thinking this gitignore caused the libs and header file to be skipped during the copy. We'll know if it works shortly!

markphelps:
DIdnt work.. still looking

sylwang:
Hi @markphelps ! Thank you for responding so quickly. Really appreciate that you're looking into this. We will stay tuned!

sylwang:
I just tried to use the latest version v0.4.4 using `go get`
```
sy.wang@F2L0GF3HWP control-plane % go get "go.flipt.io/flipt-client"
go: downloading go.flipt.io/flipt-client v0.4.4
go: added go.flipt.io/flipt-client v0.4.4
```
and still saw the same issue. 
```
[...]

PR #190, opened 2024-03-26
fix: use committed c-header file and copy to correct place expected by Go SDK

markphelps:
Attempt to fix https://github.com/flipt-io/flipt-client-sdks/issues/187

This removes the usage of `cbindgen` temporarily to just use a committed C header file `flipt_engine.h` that is required by the Go SDK.

It also copies the `flipt_engine.h` to the correct path expected by the SDK
I just went ahead and copied it to a similar path for the non-Go SDKs just for completeness

Also, it fixes the issue where the library files were not copied correctly to the resulting Go SDK because of the user of `index-filter` instead of `tree-filter`: https://git-scm.com/docs/git-filter-branch#Documentation/git-filter-branch.txt---tree-filterltcommandgt


GeorgeMac (approved):
👍 lets give it a go

Issue #148, opened 2024-02-21
[FLI-861] Try to handle panics in engine


markphelps:
> > ```
> > thread '<unnamed>' panicked at flipt-engine-ffi/src/lib.rs:212:79:
> > ```
> 
> Thank you @markphelps, that helps to understand the issue better. That `unwrap` could panic in many cases if there is an connection or configuration issue. It would be great if `initialize_engine` could return an error to upstream instead of panic. Don't know if that is possible with FFI.

Yeah agreed. We should see if we can recover from `panic` in FFI/Flipt Evaluation layer and always return an error instead of crashing the client. Will create an issue for this in the `flipt-client-sdk` repo

_Originally posted by @markphelps in https://github.com/flipt-io/flipt/issues/2791#issuecomment-1956875106_
            

<sub>[FLI-861](https://linear.app/flipt/issue/FLI-861/try-to-handle-panics-in-engine)</sub>

markphelps:
This from the stdlib may be helpful per the Rust FFI docs: https://doc.rust-lang.org/nomicon/ffi.html#catching-panic-preemptively

reinainblood:
@markphelps i'd like to take a stab at this one, could you assign it to me please?

ian-a-anderson:
@markphelps Any chance this would cause the locally cached Flipt data to be dropped in the event the server becomes inaccessible? The docs state that once the data has been loaded on the initial request, it should be cached and then updated based on background polling. When testing this we see the local Flipt client evaluation working continuing to succeed as expected if we disconnect the server, but this only lasts up until the next polling window. At that point the local cache seems to clear and then no evaluated flags are returned. Trying to sort out if that is expected or a bug and if the latter whether or not a new issue should be filed to cover that scenario. I would have expected the local data cache to persist forever, and only change on a successful poll against the server for new evaluation data.

markphelps:
@ian-a-anderson that shouldn't be caused by this issue. What SDK and version are you running when seeing the dropping of the local cache? We recently fixed this (about a month ago) so that it should keep the last successful state even in the event of errors from the server.

markphelps:
Actually after looking at the code, you're correct @ian-a-anderson. This should be a new issue I think.

https://github.com/flipt-io/flipt-client-sdks/blob/c2b2b401076de7a8f07fe827f5cc64d1363a0772/flipt-engine-ffi/src/evaluator.rs#L40-L42

[...]

PR #204, opened 2024-03-27
fix(ffi): prevent fatal error if there is the connection error during client creation

erka:
Sooner or later this should be fixed. 
```
thread '<unnamed>' panicked at flipt-engine-ffi/src/lib.rs:205:74:
called `Result::unwrap()` on an `Err` value: Server("failed to make request: error sending request for url (http://localhost:8080/internal/v1/evaluation/snapshot/namespace/default)")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 5
SIGABRT: abort
```

The idea is to start with empty state and use `replace_snapshot` logic to fill it.  If the service isn't available the error will be returned until the service is live and state is updated.


markphelps (approved):
great idea

erka:
related to #148

PR #215, opened 2024-03-29
fix(evaluation): Evaluation bug when mixing boolean and other constraints

erka:
related flipt-io/flipt#2910 flipt-io/flipt#2918


markphelps (approved):
I was just looking at how to do this!! thank you @erka

markphelps:
Im going to wait until Monday to release both of these at the same time I think, this and https://github.com/flipt-io/flipt/issues/2910

Issue #94, opened 2024-01-13
[FLI-768] Wasm Client SDK


markphelps:
### Language

WASM

### FFI Support

- [X] I've validated that the language supports Foreign Function Interface (FFI) bindings

### Additional Context

Now that we separated the evaluation logic into a library out from the FFI code (#91) we should now be able to add a new WASM client to allow us to then perform client side eval in the browser.

https://rustwasm.github.io/book/what-is-webassembly.html

### Willing to Implement

- [X] I'm willing to help implement the Flipt Client SDK in this language if it's accepted. 🙌

### Willing to Review

- [X] I'm willing to review contributions to the Flipt Client SDK in this language if it's accepted.

### Additional Notes

We’ll also need to figure out how to package the Wasm lib into a JS/Typescript browser SDK so we can call it from user’s browsers (ie to fix #84) 

<sub>[FLI-768](https://linear.app/flipt/issue/FLI-768/wasm-client-sdk)</sub>

PR #217, opened 2024-04-01
feat: wasm impl

### LINKED ISSUE #84: [Bug]: Error: Can't resolve 'os' with Storybook v7
### Bug Description

Trying to use the JS SDK as documented here: https://github.com/flipt-io/flipt-client-sdks/tree/main/flipt-client-node

I'm getting a JS error:

```
ERROR in ../../../../node_modules/@flipt-io/flipt-client/dist/index.js 6:11-24
Module not found: Error: Can't resolve 'os' in '/Users/Box/node_modules/@flipt-io/flipt-client/dist'
BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.
If you want to include a polyfill, you need to:
	- add a fallback 'resolve.fallback: { "os": require.resolve("os-browserify/browser") }'
	- install 'os-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
	resolve.fallback: { "os": false }
```

### Version Info

v1.34.0

### Search

- [X] I searched for other open and closed issues before opening this

### Steps to Reproduce

1. Install `"@flipt-io/flipt-client": "0.0.3",`

2. Setup:
```
import { FliptEvaluationClient } from '@flipt-io/flipt-client';

// namespace is the first positional argument and is optional here and will have a value of "default" if not specified.
// engine_opts is the second positional argument and is also optional, the structure is:
// {
//  "url": "http://localhost:8080",
//  "update_interval": 120,
//  "auth_token": "secret"
// }
//
// You can replace the url with where your upstream Flipt instance points to, the update interval for how long you are willing
// to wait for updated flag state, and the auth token if your Flipt instance requires it.
const fliptEvaluationClient = new FliptEvaluationClient();

const variant = fliptEvaluationClient.evaluateVariant("flag1", "someentity", {"fizz": "buzz"});

console.log(variant);
```

3. Start Storybook on the app, see error.

### Expected Behavior

App loads

### Additional Context

_No response_

markphelps:
Re: #94 

Creates `flipt-client-wasm` engine that we can use in a new browser capable SDK that will resolve eventually #84 

~~Still need to package it correctly and implement the browser capable SDK that will resolve the Flipt state and pass it to the WASM engine~~

Im going to open this up for review now as the changeset is already large enough I think.

In another PR I plan to:

- Document the `flipt-client-browser` sdk
- Add more tests
- Add tests for the `flipt-engine-wasm` in dagger
- Figure out publishing/release flow for both `flipt-engine-wasm` and `flipt-client-browser`

## TODO

- [x] create flipt-browser SDK to consume this library
- [x] publish [flipt-engine-wasm ](https://www.npmjs.com/package/@flipt-io/flipt-engine-wasm) library 
- [x] add ITs to our dagger IT suite
- [ ] publish flipt-browser SDK


ARCHITECTURE.md
```diff
@@ -1,5 +1,14 @@
 # Architecture
 
+This document describes the architecture of the client-side SDKs for Flipt.
+
+The Flipt client-side SDKs are currently broken down into two main implementations:
+
+- [FFI](#ffi)
+- [WebAssembly](#webassembly)
+
+You're use case will determine which implementation is best for you. The FFI implementation is designed to be embedded in native language client SDKs, while the WebAssembly implementation is designed to be embedded in client-side SDKs that run in the browser.
```
GeorgeMac, line 10:
```suggestion
You're use case will determine which implementation is best for you. The FFI implementation is designed to be embedded in native language client SDKs, while the WebAssembly implementation is designed to be embedded in client-side SDKs that runs in the browser.
```

flipt-client-browser/src/models.ts
```diff
@@ -0,0 +1,63 @@
+export type Authentication = string;
+
+export function WithClientTokenAuthentication(client_token: string) {
+  return `Bearer ${client_token}`;
+}
+
+export function WithJWTAuthentication(jwt_token: string) {
+  return `JWT ${jwt_token}`;
+}
```
GeorgeMac, line 9: Is it more common to lower-case these as they're functions?
Im unsure what makes most sense to folks, just looked quite like our Go one and I wondered if that is expected in TS land.

flipt-engine-wasm/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        self.store = store;
+    }
+
+    pub fn evaluate_boolean(&self, request: JsValue) -> Result<JsValue, JsValue> {
+        let req: fliptevaluation::EvaluationRequest = match serde_wasm_bindgen::from_value(request)
+        {
+            Ok(r) => r,
+            Err(e) => {
+                panic!("Invalid JSON: {}", e);
```
GeorgeMac, line 101: More of an out of my own interest question: Is the panicing here just a while we're prototyping thing, or is this what the wasm bindgen lib kind of expects you to do r.e. error handling?

GeorgeMac (approved):
Looks great man. Couple questions mostly for my own clarification.

flipt-client-browser/src/models.ts
markphelps: yeah, tbh i dont love this. I might try and make it like our existing `flipt-client-node` auth where its simply an object config like:

```
"authentication": {
  "client_token": "secret"
}
```

flipt-engine-wasm/src/lib.rs
markphelps: yeah im not sure tbh. we're using [console_error_panic_hook](https://github.com/rustwasm/console_error_panic_hook) lib here to capture all panics and show them in the JS / browser console.

I don't know if we can return actual Errors from Rust and have the same behavior. From [this thread](https://www.reddit.com/r/rust/comments/17hfljz/current_practice_for_return_errors_from/) it seems like it might be possible but with more work?

PR #219, opened 2024-04-03
chore: Package wasm

markphelps:
This should allow us to automatically publish the `flipt-engine-wasm` wasm Node module via Release Please

It should also publish the `flipt-client-browser` SDK once tagged and pushed like the other sdks


GeorgeMac (approved):
Schweet

PR #223, opened 2024-04-04
chore: update readmes and fmt with prettier

markphelps:
minor updates mostly to READMEs so that they make more sense if they are used as the front page on the registries, such as https://www.npmjs.com/package/@flipt-io/flipt-client-browser


GeorgeMac (approved):
Spring clean 🧹

PR #226, opened 2024-04-04
feat(wip): consolidate browser and wasm engine

markphelps:
Redoes our `flipt-client-browser` sdk to contain the `flipt-engine-wasm` code

After spending all day on this, this is the only way I could actually get it to work. Also, I don't really think it makes any sense to publish the `flipt-engine-wasm` lib as it's own NPM package since the `browser-client` is the only thing that can use it currently. Maybe this will change in the future

This was inspired by this blog post: https://rybicki.io/blog/2023/06/27/rust-crate-into-typescript-library.html

I spent a lot of time today trying to get it to work the way wasm-pack [mentions](https://rustwasm.github.io/wasm-pack/book/tutorials/npm-browser-packages/using-your-library.html), but it didn't work as expected since we want this to run in a browser, not in a node env.

Also I realized I very much dislike Webpack.

I tested this out by moving our NextJS example in the main repo to use this client instead of the node server side SDK: https://github.com/flipt-io/flipt/compare/nextjs-clientside (still need to clean this up some)

Here you can see it making the requests using WASM

CleanShot 2024-04-04 at 19 30 54@2x

## TODO

- [ ] Add back tests
- [x] Publish/ Fix workflows


GeorgeMac (approved):
I see you're having fun with this one.

PR #232, opened 2024-04-07
feat: Fix wasm

markphelps:
Fixes: #94

Finally get flipt-client-browser SDK to work for both commonJS and ESM using https://rollupjs.org/

This also moves the WASM code back to `flipt-engine-wasm` as I think we can also leverage this to redo the `flipt-client-node` SDK to remove FFI and swap to wasm

See: https://github.com/flipt-io/flipt/pull/2952

## TODO

- [ ] Still need to get the browser tests working, having trouble with Jest. But will do in another PR


markphelps:
Cc @erka

flipt-client-browser/src/index.ts
```diff
@@ -1,4 +1,4 @@
-import init, { Engine } from '../pkg';
+import init, { Engine } from '../dist/flipt_engine_wasm.js';
```
GeorgeMac, line 1: Was this the pesky fix?

GeorgeMac (approved):
Nice one guys!

flipt-client-browser/src/index.ts
markphelps: Yeah pretty much

PR #268, opened 2024-05-15
chore(evaluation): make clippy happy about clone

erka:
https://github.com/flipt-io/flipt-client-sdks/actions/runs/9067151090/job/24911610703


markphelps (approved):
thank you!

PR #312, opened 2024-07-29
fix: always return default reason/variant if nothing else

markphelps:
Re: https://github.com/flipt-io/flipt/pull/3313


erka (approved):
nice!

PR #316, opened 2024-08-01
chore: flatten modules to make simpler; update dependabot file

markphelps:
- Flatten some modules to the root to make jumping around a bit easier
- Update dependabot to include `flipt-client-browser` and `flipt-engine-wasm` correctly
- Move an FFI only error into the `flipt-engine-ffi` lib


erka (approved):
lgtm

Issue #141, opened 2024-02-19
[FLI-856] Support Alpine/Musl


markphelps:
Currently the engine is only compiled against GLibc which means the clients will not work in Alpine env

We should create musl compatible versions, targeting `x86_64-unknown-linux-musl` and `aarch64-unknown-linux-musl`

I think there are two paths we could take:

1. Continue to bundle these new libs along with the glibc versions so that the native clients (ie `flipt-client-java`) should 'just work' on all architectures
2. Create new musl/alpine versions of each `flipt-client-x` library, putting the responsibility on the user/integrator to select the correct SDK for their use case.

I'm leaning on 2 as it might be very difficult for us to figure out which version of the `fliptengine` Clib to 'load' at runtime when the clients initialize (ie likely not all languages allow you to know which `libc` your os supports/requires). Also each additional `fliptengine` we package increases the overall library side by avg of 20-30mb (perhaps we can find a way to slim these down though?)

I'd be interested to know what others think, also if there are any examples of other SDKs built with FFI and run on both libc and musl based OSes

<sub>From [SyncLinear.com](https://synclinear.com) | [FLI-856](https://linear.app/flipt/issue/FLI-856/support-alpinemusl)</sub>

## SDKs to Support

- [x] Go (go.flipt.io/flipt-client@musl-vx.x.x)
- [x] Node (now uses wasm)
- [ ] Python
- [ ] Ruby
- [ ] Dart
- [x] Java

markphelps:
See: https://github.com/flipt-io/flipt-client-sdks/actions/runs/7962787687/job/21736987230?pr=140#step:5:1395 for POC of failing test using `musl` based docker image

markphelps:
cc @piclem

piclemx:
@markphelps I would go with the second option also.

jalaziz:
What would option 2 look like in practice?

For example, if I import the `go` library, would I have to choose between an glibc version or a musl version? That would be somewhat problematic in scenarios where local development environments are glibc, but the production environment is musl.

jalaziz:
@markphelps Spent some time looking into this.

I think a good example repo to look at is the [confluent-kafka-go](https://github.com/confluentinc/confluent-kafka-go) library since they use the `librdkafka` C-library under the hood.

It looks like they use Go build tags to support musl vs libc, which seems like a better option than different imports.

[...]

PR #326, opened 2024-08-09
snapshot engine in go

anthonpetrow:
In our company, we tried to get involved in the hot method for a batch request, and our request began to be answered by clients with a timeout error. We carefully studied the query logic inside the repository and it turns out that inside the batch evaluates each flag in the loop and executes a query to the database each time, which increased our query execution time. We started thinking towards flipt-client-side, but we were afraid to include the ffi utility in the legacy php project. As a result, we decided to completely import the evaluation logic into our project in the go lang. Yes, it duplicates the logic of the main server, but I want to share the work done. Perhaps someone will find it useful


markphelps:
Thanks for opening this PR @anthonpetrow !!

I think this is a great reference for anyone looking to do the same in their language of choice.

> we were afraid to include the ffi utility in the legacy php project. 

We've been thinking about this recently, particularly because FFI can be a bit daunting and doesn't work in all environments (at least the way that our client side evaluator is currently built, ex: #141 ).

One approach we are actively investigating is leaning further into WASM for the evaluation logic. Then using WASM runtimes in each of the languages that support it (Go is one) to call the evaluation functions.

Currently though, because WASM / [WASI](https://wasi.dev/) spec is not complete, and definitely not supported everywhere, we will have still perform the HTTP requests in the native language like we do in [flipt-client-browser](https://github.com/flipt-io/flipt-client-sdks/blob/af8bca5d22d7b035d9bd229f9de59d2ec08d7716/flipt-client-browser/src/index.ts#L73) and then pass the data to the wasm engine as JSON.

I think we could use a lot of what you have done in this PR, but instead of re-implementing the evaluation logic in pure Go we could call into our WASM engine (after adding WASM runtime in Go also).

This would make the Go SDK cross-platform and we could then expand to other languages as well.

I'm going to open an issue to track native WASM support to get started, and will try to hack on this some this weekend.

Issue #327, opened 2024-08-09
[FLI-1159] Experiment: Go + WASM


markphelps:
There have been several issues with using FFI to call from host languages into the FFI engine.

See:

* #141
* #326

Ex:

> we were afraid to include the ffi utility in the legacy php project.

We've been thinking about this recently, particularly because FFI can be a bit daunting and doesn't work in all environments.

One approach we are actively investigating is leaning further into WASM for the evaluation logic. Then using WASM runtimes in each of the languages that support it (Go is one) to call the evaluation functions.

Currently though, because WASM / [WASI](https://wasi.dev/) spec is not complete, and definitely not supported everywhere, we will have still perform the HTTP requests in the native language like we do in [flipt-client-browser](https://github.com/flipt-io/flipt-client-sdks/blob/af8bca5d22d7b035d9bd229f9de59d2ec08d7716/flipt-client-browser/src/index.ts#L73) and then pass the data to the wasm engine as JSON.

I think we could leverage some of what was done in #326 (Go) to fetch the state like we do in `flipt-client-browser` but instead of re-implementing the evaluation logic in pure Go we could call into our WASM engine (after adding WASM runtime in Go also).

This would make the Go SDK cross-platform and we could then expand to other languages as well.

This issue tracks the progress/TODOs for trying this in Go first as we know Go the best internally.

## TODO

- [ ] Refactor evaluation lib to remove any HTTP bits (maybe continue on with [https://github.com/flipt-io/flipt-client-sdks/compare/main...rust](https://github.com/flipt-io/flipt-client-sdks/compare/main...rust))
- [ ] Build a native WASM engine (no wasmbindgen as it only works for JS)
- [ ] Create a version of the Go SDK that pulls in a WASM runtime to call into the wasm eval function
- [ ] Do all HTTP calls in the Go SDK
- [ ] Test, Package, Ship

<sub>From [SyncLinear.com](https://synclinear.com) | [FLI-1159](https://linear.app/flipt/issue/FLI-1159/experiment-go-wasm)</sub>

PR #328, opened 2024-08-14
feat: wasm non-js

markphelps:
Re: #327 

POC to try using plain wasm / wasmtime to do evaluation without FFI

will fill out more details later


GeorgeMac:
Very excited to see us explore WASM more for this.
Just a note though (you're probably aware): wasmtime-go is CGO and C FFI. So doesn't address the problem of having to have wasmtime installed for the host OS in order to use these clients. As opposed to say wazero which can be entirely statically compiled into the client.

markphelps:
> Very excited to see us explore WASM more for this. Just a note though (you're probably aware): wasmtime-go is CGO and C FFI. So doesn't address the problem of having to have wasmtime installed for the host OS in order to use these clients. As opposed to say wazero which can be entirely statically compiled into the client.

yeah we could try wazero as well instead of wasmtime. another potential option is use an approach like mentioned here https://github.com/flipt-io/flipt-client-sdks/issues/141#issuecomment-2016559343, re libkafka maybe? would the end user still need to install wasmtime locally then?

PR #332, opened 2024-08-23
feat: musl build of ffi engine

markphelps:
Re: #141 

- build flipt-engine-ffi with musl so it can run on alpine hosts
- ~~remove rust-cross (uses docker to cross compile) and do it 'manually' on the build machines~~
- moves downloading of build ffi engines into dagger / Go when packaging the SDKs (easier to express in code than YAML)

trying this with the Go SDK first, it will create a new tag `vX.X.X-musl` for the Go SDK which will contain only the musl libs

## Testing

After getting this to work (with using cross again), I tested this by creating a Dockerfile locally, and putting the built lib from the GitHub actions workflow in the correct place (`flipt-client-go/ext/linux_arm64` as im on a m1 mac)

```Dockerfile
FROM golang:1.21.3-alpine

RUN apk add --no-cache build-base

WORKDIR /src

COPY . .

ENV FLIPT_URL="https://try.flipt.io"
ENV FLIPT_AUTH_TOKEN="flipt"
ENV LD_LIBRARY_PATH="/src/ext/linux_arm64:$LD_LIBRARY_PATH"

RUN go mod download

RUN go test ./...
```

It runs correctly, just fails since the data at `https://try.flipt.io` doesnt exist

```
 > [6/6] RUN go test ./...:                                                                                                                                             
5.069 --- FAIL: TestVariant (0.00s)                                                                                                                                     
5.069     evaluation_test.go:39:                                                                                                                                        
5.069         	Error Trace:	/src/evaluation_test.go:39                                                                                                              
5.069         	Error:      	Not equal:                                                                                                                              
5.069         	           	expected: "success"
5.069         	           	actual  : "failure"
5.069         	           	
5.069         	           	Diff:
5.069         	           	--- Expected
5.069         	           	+++ Actual
5.069         	           	@@ -1 +1 @@
5.069         	           	-success
5.069         	           	+failure
5.069         	Test:       	TestVariant
5.069     evaluation_test.go:40: 
5.069         	Error Trace:	/src/evaluation_test.go:40
5.069         	Error:      	Should be empty, but was invalid request: failed to get flag information default/flag1
5.069         	Test:       	TestVariant
```


.github/workflows/package-ffi-engine.yml
```diff
@@ -54,12 +61,33 @@ jobs:
...
+          if [[ "${{ matrix.platform.os }}" =~ ^ubuntu.*$ ]]; then
+            sudo apt-get install -qq musl-tools musl-dev crossbuild-essential-arm64
+          fi
+
+          # some additional configuration for cross-compilation on linux
+          cat >>~/.cargo/config <<EOF
+          [target.aarch64-unknown-linux-gnu]
+          linker = "aarch64-linux-gnu-gcc"
+          [target.aarch64-unknown-linux-musl]
+          linker = "aarch64-linux-gnu-gcc"
```
erka, line 77: is it correct?

markphelps: Not sure, it was required to get the engine compiled with openssl support for musl

I got the example from https://github.com/nicolas-van/rust-cross-compile-example/blob/main/.github/workflows/rust.yml#L83-L88

I now realize though that in their example they built a binary, and we are building a library so this may not actually work once its called from the host machine (ie alpine) if it is still linked against glibc

erka: AWS uses it in the [docs](https://github.com/awslabs/aws-lambda-web-adapter/blob/5d7885bd26b3b838e0884865f301ab01e14d8960/docs/development.md#compiling-on-macos) like 
```toml
[target.aarch64-unknown-linux-musl] 
linker = "aarch64-unknown-linux-musl-gcc"
```

markphelps: yeah let me try that.

i downloaded the engine built for musl and it does seem like its linked against glibc, which wont work I believe on alpine:

CleanShot 2024-08-24 at 13 15 57

markphelps: It cant seem to find `aarch64-linux-musl-gcc` pr ` aarch64-unknown-linux-musl-gcc` (I tried both)

https://github.com/flipt-io/flipt-client-sdks/actions/runs/10540369969/job/29204942815#step:6:305

I think `aarch64-unknown-*` might be specific to MacOS/Homebrew per https://github.com/awslabs/aws-lambda-web-adapter/blob/5d7885bd26b3b838e0884865f301ab01e14d8960/docs/development.md#compiling-on-macos

Also it seems others recommend setting  the linker to `aarch64-linux-gnu-gcc` 🤷🏻 

https://github.com/rust-lang/stacker/issues/80#issuecomment-1547991131

erka: Could this warning have an effect on it?
```
warning: openssl-sys@0.9.103: Compiler family detection failed due to error: ToolNotFound: Failed to find tool. Is `aarch64-linux-musl-gcc` installed?
```
I see that someone installed it before the build in the [action](https://github.com/sxyazi/yazi/pull/759/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R25-R27)

erka: > yeah let me try that.
> 
> i downloaded the engine built for musl and it does seem like its linked against glibc, which wont work I believe on alpine:
> 
> CleanShot 2024-08-24 at 13 15 57

maybe it will work.

markphelps: I switched it back, it builds now again without any warning: https://github.com/flipt-io/flipt-client-sdks/actions/runs/10542101107/job/29208497781

I havent yet tested it in an actual alpine container yet though.. trying to get that setup now

markphelps: It does not like that:

```
 > [6/6] RUN go test ./...:                                                                                                                                             
4.814 FAIL	go.flipt.io/flipt-client [build failed]                                                                                                                 
4.814 FAIL                                                                                                                                                              
4.814 # go.flipt.io/flipt-client.test                                                                                                                                   
4.814 /usr/local/go/pkg/tool/linux_arm64/link: running gcc failed: exit status 1                                                                                        
4.814 /usr/lib/gcc/aarch64-alpine-linux-musl/12.2.1/../../../../aarch64-alpine-linux-musl/bin/ld: warning: libc.so.6, needed by /src/ext/linux_arm64/libfliptengine.so, not found (try using -rpath or -rpath-link)
4.814 /usr/lib/gcc/aarch64-alpine-linux-musl/12.2.1/../../../../aarch64-alpine-linux-musl/bin/ld: warning: ld-linux-aarch64.so.1, needed by /src/ext/linux_arm64/libfliptengine.so, not found (try using -rpath or -rpath-link)
```

erka (approved):
I tried the artifact on x86_64 alpine and works great.

PR #341, opened 2024-08-26
chore: make evaluation lib code and evaluation response types more idiomatic

markphelps:
brought to you by AI

this will actually be a breaking change for Java and Python SDKs as it changes `variant_key` and `variant_attachment` fields to be Optional 

I think we should use Optional where relevant for those languages that support it

### DIFF
--- a/flipt-client-java/src/main/java/io/flipt/client/models/VariantEvaluationResponse.java
+++ b/flipt-client-java/src/main/java/io/flipt/client/models/VariantEvaluationResponse.java
@@ -37,8 +37,8 @@ public class VariantEvaluationResponse {
     this.segmentKeys = segmentKeys;
     this.reason = reason;
     this.flagKey = flagKey;
-    this.variantKey = variantKey;
-    this.variantAttachment = variantAttachment;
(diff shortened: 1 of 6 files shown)


GeorgeMac (approved):
Thank you robots

erka:
Will this also conflict with the GRPC schema of the Flipt API? Those fields are `string` without `optional` modifier. wdyt?

markphelps:
I'm also OK with not doing the optional changes right now though if we think it would make for an awkward experience / upgrade

erka:
> I'm also OK with not doing the optional changes right now though if we think it would make for an awkward experience / upgrade

Maybe the `optional` is correct thing but it creates inconsistency in the  Flipt API for current version imho.

markphelps:
> > I'm also OK with not doing the optional changes right now though if we think it would make for an awkward experience / upgrade
> 
> Maybe the `optional` is correct thing but it creates inconsistency in the Flipt API for current version imho.

Yeah I can revert those bits and just leave a comment that we should/could make them optional in the future but would be breaking API

erka (approved):

PR #342, opened 2024-09-01
docs(browser): Add documentation around custom fetcher and etag support

markphelps:
Add docs for browser sdk around custom fetcher and etag support


erka (approved):
nice!

PR #349, opened 2024-09-03
fix(go): return errors correctly from Go client

markphelps:
Trying to test invalid authentication errors from #347

### DIFF
--- a/flipt-client-go/evaluation_test.go
+++ b/flipt-client-go/evaluation_test.go
@@ -37,13 +37,10 @@ func init() {
 func TestInvalidAuthentication(t *testing.T) {
 	client, err := flipt.NewClient(flipt.WithURL(fliptUrl), flipt.WithClientTokenAuthentication("invalid"))
 	require.NoError(t, err)
-	variant, err := client.EvaluateVariant(context.TODO(), "flag1", "someentity", map[string]string{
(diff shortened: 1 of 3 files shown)


flipt-client-go/evaluation_test.go
```diff
@@ -30,6 +34,11 @@ func init() {
 	}
 }
 
+func TestInvalidAuthentication(t *testing.T) {
+	_, err := flipt.NewClient(flipt.WithURL(fliptUrl), flipt.WithClientTokenAuthentication("invalid"))
+	assert.EqualError(t, err, "invalid authentication")
```
markphelps, line 39: i know this isnt the correct error string, but i just wanted to see if there is any error

erka: we need to do some `variant, err := evaluationClient.EvaluateVariant(...)` to get the error.

markphelps: it should return the error on the initial pulling of the state I would think. I updated the test and still no error, maybe we arent returning it correctly from the engine?

erka: let me give it a try.

erka (approved):
nice. I believe this actually the fix for original issue. ty!

PR #350, opened 2024-09-03
refactor(go)!: hide internals of ffi communication and provide cleaner client

erka:
This is a breaking change on API level. 
As it's still v0 I think it's okay to make this change

### DIFF
--- a/flipt-client-go/evaluation.go
+++ b/flipt-client-go/evaluation.go
@@ -14,8 +14,11 @@ import "C"
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"unsafe"
 )
+
+const statusSuccess = "success"
 
 // Client wraps the functionality of making variant and boolean evaluation of Flipt feature flags.
 type Client struct {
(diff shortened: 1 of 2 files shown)


erka:
@markphelps this is PR into your pull request

flipt-client-go/evaluation.go
```diff
@@ -157,11 +163,15 @@ func (e *Client) EvaluateBoolean(_ context.Context, flagKey, entityID string, ev
 		return nil, err
 	}
 
-	return br, nil
+	if br.Status == statusSuccess {
```
markphelps, line 166: we could make this a non breaking API change by returning `br` here and still check if `br.Status == success` and return an error if not right?

i just worry about making this breaking change in only one sdk until we have tried it in all client side sdks to make sure it is the 'shape' of the responses that we want.

wdyt?

markphelps (commented):
thank you! I do have a question on if we should make this breaking change now or do another pass of all the client side sdk's to make sure we are somewhat consistent if we are going to change the response types

flipt-client-go/evaluation.go
erka: yes, we can do that.  Please revert those or I will do it tomorrow.

Issue #32, opened 2023-12-18
[FLI-770] need sdk flutter


baseron01:
null

<sub>[FLI-770](https://linear.app/flipt/issue/FLI-770/need-sdk-flutter)</sub>

markphelps:
@baseron01 would you be interested in helping contribute? We do not have Dart/Flutter experience as a team

markphelps:
@baseron01 our v0.1.0 for our Dart/Flutter SDK was just released: https://pub.dev/packages/flipt_client/versions/0.1.0 🎉 

It only supports Mac / Linux for now, iOS and Android coming next!

PR #351, opened 2024-09-06
feat: add dart client

markphelps:
Re: #32

Adds `flipt-client-dart` SDK for dart language using FFI

## TODO

- [x] Add batch support
- [x] Add list flags support
- [x] Check if idiomatic
- [x] ~~Publish/Release~~ will do in another PR

### DIFF
--- a/flipt-client-dart/lib/src/flipt_evaluation_client.dart
+++ b/flipt-client-dart/lib/src/flipt_evaluation_client.dart
@@ -91,7 +91,7 @@ class FliptEvaluationClient {
     }
   }
 
-  Result<VariantEvaluationResponse> evaluateVariant(
+  VariantEvaluationResponse evaluateVariant(
       {required String flagKey,
       required String entityId,
       required Map<String, dynamic> context}) {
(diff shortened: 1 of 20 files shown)


flipt-client-dart/lib/src/flipt_evaluation_client.dart
```diff
@@ -0,0 +1,231 @@
...
+      entityId: entityId,
+      context: context,
+    );
+
+    final requestJson = jsonEncode(request.toJson());
+    final requestUtf8 = requestJson.toNativeUtf8();
+
+    final resultPtr = evaluateVariant(_engine, requestUtf8);
+    final result = resultPtr.toDartString();
+    _destroyString(resultPtr);
```
erka, line 113: Just not critical thing but still it would be nice to have the same pattern between `evaluateBoolean`, `evaluateBatch` and `listFlags`. This section is out of try/catch block and `evaluateBoolean` has it inside. `listFlags` doesn't have try/catch block. Also I think we could call `calloc.free(requestUtf8);` much early.  Probably next to `_destroyString`.

erka (commented):
@markphelps it's huge!

I have one minor comment. I also think we should not return `Result` as `result.result` looks a bit weird. I know we follow the same pattern as other clients but it's a new client maybe it's a good time to do the break.

markphelps:
> @markphelps it's huge!
> 
> I have one minor comment. I also think we should not return `Result` as `result.result` looks a bit weird. I know we follow the same pattern as other clients but it's a new client maybe it's a good time to do the break.

Yeah I think you are right. I made those changes. I'll create an issue to make the changes in the other SDKs as well

flipt-client-dart/lib/src/flipt_evaluation_client.dart
markphelps: yeah i just removed the try / catch and called `calloc.free` right after `_destroyString`

erka (approved):
niceee

Issue #83, opened 2024-01-04
Rust Client Side SDK


markphelps:
### Language

Rust :crab:

### FFI Support

- [X] I've validated that the language supports Foreign Function Interface (FFI) bindings

### Additional Context

I dont think we'll need to go through FFI since our engine is written in Rust

### Willing to Implement

- [X] I'm willing to help implement the Flipt Client SDK in this language if it's accepted. 🙌

### Willing to Review

- [X] I'm willing to review contributions to the Flipt Client SDK in this language if it's accepted.

andrew-scott-fischer:
@markphelps What prevented this from being worked on? are you open to PRs to add it?

markphelps:
> [@markphelps](https://github.com/markphelps) What prevented this from being worked on? are you open to PRs to add it?

Only our lack of Rust library experience. Def open to PRs!!

andrew-scott-fischer:
I'll propose this as a starting point. Do you have any concerns with assuming a `tokio` runtime and using `rustls` rather than `native-tls`?

I considered making it runtime independent and allowing clients to configure how they wanted to do mtls, but I thought it might be a bit over engineered.

I thought it reasonable to make the system async by default, but provide a `blocking` feature so sync systems could still leverage the client, and that didn't seem like much of a lift.

I'd like some feedback on what constraints you think are important before I begin proposing PRs.

---

# Native Rust Client SDK Design

## Goal

Add a publishable, user-facing Rust client to `flipt-client-sdks` with a simple Tokio-first API, while preserving the existing FFI and WASM SDKs.

## Primary decision

Use Tokio and Reqwest directly in v1. Do not build a runtime-neutral abstraction in the first implementation.

The normal async path assumes the application is running a Tokio runtime:

```rust
let client = Client::builder()
    .url("https://flipt.example.com")
    .namespace("default")
    .build()
    .await?;

let result = client.evaluate_boolean(&request)?;
```

The client does not construct a hidden runtime or call `block_on` in the async path. It uses the caller's Tokio runtime for background updates.

The `blocking` feature provides a synchronous facade that explicitly owns a Tokio runtime.

## Architecture

```mermaid
flowchart TD
    Eval["flipt-evaluation<br/>Evaluation and models"]
    Core["flipt-client-core<br/>Tokio transport and state"]
    Client["flipt-client-rust<br/>Public Rust API"]
    Blocking["BlockingClient<br/>Optional feature"]
[...]

PR #357, opened 2024-09-08
refactor: remove common module

markphelps:
moves previously `common` model code into `flipt` module that is pub

this will be useful for implementing #83


erka (approved):
nice

PR #372, opened 2024-09-10
chore(ffi): replace openssl with rust-native-tls crate

erka:
`reqwest` uses `rust-native-tls` for tls. It requires openssl for linux but nothing for Windows or macOS. Per their suggestion for linux native-tls-vendored feature should be used to compile a copy of OpenSSL. This change reduces compilation time during the development on macOS


erka:
> nice! im assuming this will work on mobile too? re: #363

Let's merge that first and check it in this PR

markphelps:
> > nice! im assuming this will work on mobile too? re: #363
> 
> Let's merge that first and check it in this PR

Mind giving me a ✅ on it?

erka:
> Mind giving me a ✅ on it?

Nope, give it a try

markphelps (approved):
Good find!

Issue #353, opened 2024-09-07
[FLI-1269] Ensure consistency between SDKs


markphelps:
I think we should audit our existing client SDKs in the various languages and ensure they have a similar API before we make any of them 1.0.

Heres a few areas I think we could check for consistency/improve upon:

1. Remove the `Result` types as public facing return types. Similar to how we now do in the new dart SDK ([https://github.com/flipt-io/flipt-client-sdks/pull/351#pullrequestreview-2287892390](https://github.com/flipt-io/flipt-client-sdks/pull/351#pullrequestreview-2287892390)), just return the underlying types instead of the wrapped `Result` types
2. Ensure consistent naming of Request/Response types. ie: `EvaluationRequest`, `EvaluationResponse`, `BooleanEvaluationResponse`, `VariantEvaluationResponse`, etc.
3. Ensure the methods are named consistently across SDKs (ie: `evaluateVariant` / `evaluate_variant` depending on language idioms, `evaluateBoolean`, `evaluateBatch`, `listFlags`)
4. Ensure all methods are supported in all SDKs. I think we may be missing `listFlags` in a few
5. Ensure clients are named similarly, ex: `flipt-client` instead of `flipt-client-{language}` for their published names. As each ecosystem will have its own package manager, so the `-{language}` bit is unnecessary for the public facing package name (where applicable, ie we still need `flipt-client` and `flipt-client-browser` as they are both NPM packages)
6. Ensure types/classes named consistently, ie: `FliptEvaluationClient` instead of `FliptClient` or maybe vice-versa?
7. Dont expose anything publicly in the API that references `Engine`, ie `EngineOpts/Options`. The fact that we call the Rust evaluator an engine under the hood should not leak externally
8. Use `optional` types for languages that support it. ie: `variantKey` and `variantAttachment` should be optional in Variant evaluation responses

Anything else? Feel free to add

<sub>From [SyncLinear.com](https://synclinear.com) | [FLI-1269](https://linear.app/flipt/issue/FLI-1269/ensure-consistency-between-sdks)</sub>

## SDKS to Refactor

- [x] Node
- [x] Browser
- [x] Go
- [x] Dart
- [x] Ruby
- [ ] Python
- [x] Java

## Published

- [x] Node
- [x] Browser
- [ ] Go
- [x] Dart
- [ ] Ruby
- [ ] Python
- [x] Java

PR #375, opened 2024-09-11
refactor!(go):  hide result types from go-sdk

erka:
related #353

### DIFF
--- a/flipt-client-go/README.md
+++ b/flipt-client-go/README.md
@@ -51,7 +51,7 @@ import (
 )
 
 func main() {
-  evaluationClient, err := flipt.NewClient()
+  evaluationClient, err := flipt.NewEvaluationClient()
   if err != nil {
     log.Fatal(err)
   }
--- a/flipt-client-go/README.md
+++ b/flipt-client-go/README.md
@@ -23,7 +23,6 @@ This SDK currently supports the following OSes/architectures:
 - MacOS x86_64
 - MacOS arm64
 
-
 ### Glibc vs Musl
 
 Most Linux distributions use [Glibc](https://en.wikipedia.org/wiki/Glibc), but some distributions like Alpine Linux use [Musl](https://en.wikipedia.org/wiki/Musl). If you are using Alpine Linux, you will need to install the `musl` tagged version of the client.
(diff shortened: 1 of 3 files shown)


erka:
>  can we rename `EngineOpts` to `ClientOptions` like in [73a1d84](https://github.com/flipt-io/flipt-client-sdks/commit/73a1d84a0f7c9c18fda4d031a84145a533a8c1a3)
> 
> ```
> engOpts := &EngineOpts[any]{
> 		URL:            client.url,
> 		UpdateInterval: client.updateInterval,
> 		Authentication: &client.authentication,
> 		Reference:      client.ref,
> 	}
> ```
> 

@markphelps I'd appreciate it if we could verify this. I'm thinking that `EngineOpts` should not be exposed publicly since it's only used internally within the system. Additionally, are go `clientOption` and node `ClientOptions` actually equivalent in terms of their definition at this specific commit?

markphelps:
> > can we rename `EngineOpts` to `ClientOptions` like in [73a1d84](https://github.com/flipt-io/flipt-client-sdks/commit/73a1d84a0f7c9c18fda4d031a84145a533a8c1a3)
> > ```
> > engOpts := &EngineOpts[any]{
> > 		URL:            client.url,
> > 		UpdateInterval: client.updateInterval,
> > 		Authentication: &client.authentication,
> > 		Reference:      client.ref,
> > 	}
> > ```
> 
> @markphelps I'd appreciate it if we could verify this. I'm thinking that `EngineOpts` should not be exposed publicly since it's only used internally within the system. Additionally, are go `clientOption` and node `ClientOptions` actually equivalent in terms of their definition at this specific commit?

Oh yeah you're right. I forgot we only expose functional options for configuring the client. 

I do think we could still rename `Client` to `FliptEvaluationClient` though, wdty?

erka:
> I do think we could still rename `Client` to `FliptEvaluationClient` though, wdty?

This is a challenging question. I believe we should not use the same name across all clients because some languages have package name in front. For example, in Go and Ruby I would prefer `EvaluationClient`, so it could be as `flipt.EvaluationClient` in Go or `Flipt::EvaluationClient` in Ruby.  I don't like double the same word twice.  wdyt?

markphelps (approved):
awesome! ty

Issue #373, opened 2024-09-10
[FLI-1275] Try WASM impl for node SDK


markphelps:
We should try switching to the WASM engine for node, since it has first class support for WASM

https://nodejs.org/en/learn/getting-started/nodejs-with-webassembly

<sub>[FLI-1275](https://linear.app/flipt/issue/FLI-1275/try-wasm-impl-for-node-sdk)</sub>

PR #378, opened 2024-09-11
feat(node): reimplement node SDK in wasm

markphelps:
Fixes: #373 

POC re-implementing the node SDK with wasm instead of FFI

This will make it platform independent!

Also changes the fields in `flipt-client-node` to be camelCase instead of snake_case which is more idiomatic for TS/JS

## TODO

- [x] Cleanup/update documentation
- [x] Implement periodic fetching in NodeJS of flag state?
- [x] Add back browser tests (need to split up the call to `wasm-pack` to allow for both `nodejs` and `web`
- [x] Update client to return `XResponse` instead of `Result` types
- [x] Update packaging workflow for Node client


flipt-client-node/README.md
```diff
@@ -66,27 +62,75 @@ The `FliptEvaluationClient` supports the following authentication strategies:
...
+```typescript
+// Refresh the flag state
+fliptEvaluationClient.refresh();
+```
 
-To work around this issue, it is recommended to run your tests with the `--pool=forks` flag.
+## ETag Support
 
-See [#169](https://github.com/flipt-io/flipt-client-sdks/issues/169) for more information.
+The default fetcher uses [ETag HTTP headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) to reduce the number of requests to the Flipt instance. This is useful in scenarios where the flag state is not frequently updated and you want to reduce the load on the server.
```
GeorgeMac, line 108: Small clarification: doesn't change the number of requests (just avoids resending previously observed state):
```suggestion
The default fetcher uses [ETag HTTP headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) to reduce overhead building and sending previously observed snapshots. This is useful in scenarios where the flag state is not frequently updated and you want to reduce the load on the server.
```

GeorgeMac (approved):
Change is looking great. One suggestion in docs.

flipt-client-node/src/index.ts
```diff
@@ -1,202 +1,265 @@
...
+          headers
+        });
+
+        if (resp.status === 304) {
+          return resp;
+        }
+
+        if (!resp.ok) {
+          throw new Error('Failed to fetch data');
+        }
```
erka, line 83: Some minor thing but could we add more details about the status in the error.  It would so easier to understand the problem if we add some `statusText` into the error message.  Also I would like double check if we want accept 200-299 range. Probably only 200 should be valid status code, right?

erka (approved):
Great work @markphelps! 

I have only one minor comment about the error message.

flipt-client-node/src/index.ts
markphelps: I added `statusText` to the errors where it can occur, im not to worked about non 2xx errors though

PR #379, opened 2024-09-12
fix(java): loading lib on arm

markphelps:
When I run `dagger run go run ./test --sdks=java` on main (on ARM machine) I get:

```
> Task :test FAILED

TestFliptEvaluationClient > initializationError FAILED
    java.lang.UnsatisfiedLinkError at TestFliptEvaluationClient.java:25
```

When I run with this fix I get a successful test passing

### DIFF
--- a/.github/workflows/package-browser-sdk.yml
+++ b/.github/workflows/package-browser-sdk.yml
@@ -8,24 +8,26 @@ permissions:
 
 env:
   NPM_API_KEY: ${{ secrets.NPM_API_KEY }}
+  GO_VERSION: "1.22"
+  DAGGER_VERSION: "0.12.3"
 
 jobs:
   build:
+    name: Build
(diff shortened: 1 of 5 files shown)


.github/workflows/package-browser-sdk.yml
```diff
@@ -8,24 +8,26 @@ permissions:
 
 env:
   NPM_API_KEY: ${{ secrets.NPM_API_KEY }}
+  GO_VERSION: "1.22"
+  DAGGER_VERSION: "0.12.3"
```
markphelps, line 12: dagger recently removed a bunch of things: https://github.com/dagger/dagger/releases/tag/v0.13.0

going to pin the version to what we have in `flipt`, maybe we could consider bumping everything to the latest `0.12.x` version though

GeorgeMac (approved):

PR #381, opened 2024-09-13
feat: add list flags to wasm/node SDK

markphelps:
- Adds list flags support to wasm engine
- Adds list flags back to node SDK

## TODO (another PR)

- Add list flags to browser SDK


erka (approved):
nice

Issue #377, opened 2024-09-11
[FLI-1277] Don't panic in WASM engine


markphelps:
When looking into #373 , I ran the tests while Flipt is not running and got this error:

```
    panicked at flipt-engine-wasm/src/lib.rs:62:17:
    Invalid JSON: Error: invalid type: string "{\"url\":\"http://localhost:8080\",\"authentication\":{\"client_token\":\"blah\"}}", expected struct Document
```

We should not panic at all in the wasm engine and handle / return errors gracefully to the caller

<sub>[FLI-1277](https://linear.app/flipt/issue/FLI-1277/dont-panic-in-wasm-engine)</sub>

PR #382, opened 2024-09-13
fix(wasm): prevent panic in the wasm

erka:
Fixes: #377

### DIFF
--- a/flipt-client-browser/src/index.ts
+++ b/flipt-client-browser/src/index.ts
@@ -100,7 +100,8 @@ export class FliptEvaluationClient {
 
     const data = await resp.json();
 
-    const engine = new Engine(namespace, data);
+    const engine = new Engine(namespace);
+    engine.snapshot(data);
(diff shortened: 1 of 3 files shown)


flipt-client-browser/src/index.ts
```diff
@@ -100,7 +100,8 @@ export class FliptEvaluationClient {
 
     const data = await resp.json();
 
-    const engine = new Engine(namespace, data);
+    const engine = new Engine(namespace);
+    engine.snapshot(data);
```
markphelps, line 104: we'll have to do the same thing in the node SDK now as well I think

erka: I missed that. Ty!

markphelps (approved):
looks great, ty!

erka:
@markphelps does kodiakhq just ignore status checks? My commit wasn't signed

markphelps:
@erka ah sorry. I think the DCO check wasn't required (i fixed it now) so kodiak merged it

PR #383, opened 2024-09-13
fix(wasm): add wasm tests

erka:
related to #382


erka:
It seems that coverage is possible, but it's currently only available on the rust nightly. It will likely require using `minicov`. Not sure if it's a priority right now.

https://github.com/rustwasm/wasm-bindgen/blob/main/guide/src/wasm-bindgen-test/coverage.md

markphelps (approved):
nice!

PR #384, opened 2024-09-14
refactor: browser argument casing + list flags

markphelps:
- redoes browser client to use similar camel casing for field/argument names like in #373 
- adds list flags method to browser SDK

## TODO

- [ ] share code between node and browser SDKs to remove duplication


erka (approved):
nice!

PR #393, opened 2024-09-22
feat: React sdk

markphelps:
Adds React SDK

See: https://github.com/flipt-io/flipt/pull/3480


flipt-client-react/README.md
```diff
@@ -0,0 +1,210 @@
...
+  if (error) return <div>Error: {error.message}</div>;
+
+  return isEnabled ? children : null;
+}
+
+function App() {
+  return (
+    <FliptProvider options={{ url: 'https://your-flipt-instance.com' }}>
+      <h1>My App</h1>
+      <FeatureFlag flagKey="new-feature" entityId="user-123">
```
GeorgeMac, line 194: Nice

GeorgeMac (approved):
This is really cool. I like the `FeatureFlag` element too. I wonder if there is value in also adding some kind of conditional rendering element for `Variant` on the variant type flag.
e.g.
```jsx
<FeatureFlag ...>
  <Variant key="one">Rendered when variant one is presented</Variant>
  <Variant key="two">Rendered when variant two is presented</Variant>
</FeatureFlag>
```

(more in a later PR)

erka:
@markphelps please don't rush to release this to public. Examples are not correct. It should be currently

```js
React.useEffect(() => {
    if (client) {
      setIsEnabled(client.evaluateBoolean("flag1", "", {}));
    }
  }, [client, flagKey, entityId]);
```

Another issue that each `useFliptContext` makes a call to Flipt. For example, if I have three `FeatureFlag` components on in App component, tree calls will be made. 

image.  

And `refresh` actually do the same thing... and it doesn't update the state of the page even if the flags are changed and update is fetched. :(

I feel that `useSyncExternalStore` should be used but I don't have nothing how to use it.

markphelps:
> @markphelps please don't rush to release this to public. Examples are not correct. It should be currently
> 
> ```js
> React.useEffect(() => {
>     if (client) {
>       setIsEnabled(client.evaluateBoolean("flag1", "", {}));
>     }
>   }, [client, flagKey, entityId]);
> ```
> 
> Another issue that each `useFliptContext` makes a call to Flipt. For example, if I have three `FeatureFlag` components on in App component, tree calls will be made.
> 
> image.
> 
> And `refresh` actually do the same thing... and it doesn't update the state of the page even if the flags are changed and update is fetched. :(
> 
> I feel that `useSyncExternalStore` should be used but I don't have nothing how to use it.

Good catches @erka ! Although since its a v0.0.1 im not too worried about it being released now with these issues, we can itterate and fix them and release newer versions.

Also im not sure if this is correct:

> 
> ```js
> React.useEffect(() => {
>     if (client) {
>       setIsEnabled(client.evaluateBoolean("flag1", "", {}));
>     }
>   }, [client, flagKey, entityId]);
> ```

since `client.evaluateBoolean(` returns a booleanResponse type, so we still need to check `enabled` on it

PR #419, opened 2024-10-03
chore(ffi): drop direct reference to tokio crate

erka:


markphelps (approved):
nice, im guessing its being pulled in transitively?

erka:
> nice, im guessing its being pulled in transitively?

Yep, FFI doesn't use `tokio` directly and `reqwest` has `tokio` as dependency and upper version.

PR #422, opened 2024-10-03
chore: update readmes re musl support for java

markphelps:


erka (approved):

erka:
nice!

erka (changes requested):
We have to change the example for java client. It's broken right now. `  FliptEvaluationClient fliptClient = null;` should be defined before `try` block

erka (approved):
Updated the example

PR #423, opened 2024-10-04
chore(java): fix package java client for musl

erka:


erka:
@markphelps Currently we got glibc `libfliptengine.so` into the musl jar. I suspect we rename the folder in a wrong way. But I  can't verify it local. Dagger is a nightmare in some way between versions.

markphelps:
> @markphelps Currently we got glibc `libfliptengine.so` into the musl jar. I suspect we rename the folder in a wrong way. But I can't verify it local. Dagger is a nightmare in some way between versions.

@erka I think i might have fixed this in my latest push to the other PR: https://github.com/flipt-io/flipt-client-sdks/pull/422/files#diff-2d2d49a7b0d105b3823a741b2e434752094d4ad843871f550d96f47d32cb4610R419

I'm trying to resolve the issue described [here](https://github.com/flipt-io/flipt-client-sdks/issues/141#issuecomment-2393160344) by also packaging in the Darwin libs as well in the musl SDKs

erka:
@markphelps Please let me know when you publish `0.0.2`. I will try it locally too.

markphelps:
> @markphelps Please let me know when you publish `0.0.2`. I will try it locally too.

It should be on mvn central by now I would think. Released it about 45 minutes ago

erka:
> It should be on mvn central by now I would think. Released it about 45 minutes ago

@markphelps yep, it's available and it works in my test app in alpine.

PR #430, opened 2024-10-08
refactor(ffi): refactor engine for streaming

markphelps:
Opening this up for feedback. It hasnt yet been tested with streaming server side changes yet, however I feel like this PR is large enough already that we should go ahead and open it.

## What

Refactors the `flipt-ffi-engine` module with the following major changes:

- Renames `HTTPParser` to `HTTPFetcher`
- Decouples fetching from the evaluation store
- Removes generic parameter `P` from evaluator as it will always use the `HTTPFetcher` (could prob do the same for `Snapshot` store in the future
- Removes unneeded method `repace_store` to just use `replace_snapshot` on state change
- Switches to use Rust [channels](https://doc.rust-lang.org/rust-by-example/std_misc/channels.html) to communicate state changes, which makes switching to leverage streaming easier as well
- Adds streaming mode, described below

## Polling / Streaming Flow

This adds initial support for https://github.com/flipt-io/flipt/issues/3458

The client will be able to be configured in either streaming or polling mode. Default will be polling, but they will be able to enable streaming via engine options (yet to be decided how this is exposed to user).

Right now this only is a refactor of the existing polling code, as the streaming functionality is never called.

## TODO

- [x] Test
- [x] Wire up ability to put in streaming mode via engine opts


flipt-engine-ffi/src/http.rs
```diff
@@ -148,207 +213,311 @@ impl Parser for HTTPParser {
...
+                    .await
+                    .map_err(|_| Error::Server("Failed to send result".into()))?;
+            }
+            Ok(None) => {}
+            Err(e) => sender
+                .send(Err(e))
+                .await
+                .map_err(|_| Error::Server("Failed to send error".into()))?,
+        }
+        Ok(())
```
GeorgeMac, line 308: Do we want this to always evaluate to `OK()` ?
Seems like this might shadow any error in the previous expresson which the caller does match on in a `let` expression to print out.

flipt-engine-ffi/src/lib.rs
```diff
@@ -97,31 +99,32 @@ impl Default for EngineOpts {
...
+        let mut rx = runtime.block_on(async { fetcher.start() });
 
-        std::thread::spawn(move || loop {
-            std::thread::sleep(std::time::Duration::from_secs(update_interval));
-            let mut lock = evaluator.lock().unwrap();
-            lock.replace_snapshot();
+        let update_handle = runtime.spawn(async move {
+            while let Some(res) = rx.recv().await {
+                evaluator_clone.lock().unwrap().replace_snapshot(res);
+            }
```
GeorgeMac, line 119: I think it is worth doing an initial replace_snapshot before spawning this thread to ensure at-least one snapshot loads. This may be the source of empty evals seen in ITs.

flipt-engine-ffi/src/http.rs
markphelps: if there's any errors they get handled in the `map_err` calls and the `?` at the end means they get returned early to the caller:

```
.map_err(|_| Error::Server("Failed to send result".into()))?;
```

flipt-engine-ffi/src/lib.rs
markphelps: that was it!

markphelps:
@GeorgeMac I think this is ready for a re-review

GeorgeMac (approved):
Nice one.

Do we want to add comments to all these new Streaming enum types to make it clear this is experimental / cloud only for now?

markphelps:
> Nice one.
> 
> Do we want to add comments to all these new Streaming enum types to make it clear this is experimental / cloud only for now?

yah will do

flipt-engine-ffi/src/http.rs
```diff
@@ -148,209 +228,376 @@ impl Parser for HTTPParser {
...
+            .map_err(|_| Error::Server("failed to send result".into()))
+    }
+
+    async fn handle_streaming(
+        &mut self,
+        sender: &mpsc::Sender<Result<source::Document, Error>>,
+    ) -> Result<(), Error> {
+        let stream_result = self.fetch_stream().await;
+
+        match stream_result {
```
erka, line 331: Maybe it's too late. Could serde_json::StreamDeserializer be used here?

markphelps: maybe, this works however. do you know if `serde_json::StreamDeserializer` would provide a benefit?

erka: I am worry about bytes and `\n` on utf-8 string. But it could be done with refactoring.

markphelps:
Im going to add more tests coverage and document the new streaming option before merging

PR #431, opened 2024-10-08
feat(browser): allow caller to know if the snapshot was updated

erka:
Sometime it would be helpful to understand if the any changes in the
snapshot after the refresh.


markphelps (approved):
great idea!

PR #434, opened 2024-10-11
fix(browser): use another way for etag checking

erka:
It turns out that fetch API will got the HTTP 200 in browser even when server response will be HTTP 304.

https://stackoverflow.com/questions/62378379/do-the-fetch-api-and-axios-see-304-responses-as-200


markphelps (approved):
nice. can you bump the browser sdk patch release version in package.json and run `npm i` as well on this PR please?

markphelps (approved):
nice ! ty

PR #450, opened 2024-10-16
feat: expose if refresh changed state in node sdk

markphelps:
Like #431 but for node


erka (approved):
it looks great

please publish browser version as well so I would test it

markphelps:
@erka Just published `@flipt-io/flipt-client-browser@0.3.1` and `@flipt-io/flipt-client@0.11.0`

PR #453, opened 2024-10-18
chore: Nextjs app router support

markphelps:
Try to work around issue re NextJS app router: https://github.com/vercel/next.js/issues/55537


erka:
@markphelps could you please address ` const interval = options.updateInterval || 0;` issue too? I believe the README says the default is `120` seconds but it's actually `0` and it uses milliseconds instead of seconds. I missed it in the last PR.

erka (approved):
niiiice!  It's great to see the support for Next.JS

PR #458, opened 2024-10-21
chore: add scripts to automate release

markphelps:
Automates releasing of these sdks instead of us having to do it manually.

The script supports the following workflows:

1. Updating the necessary files for version bumps
2. Tagging and pushing the relevant tags for each sdk
3. Both 1 + 2 combined

This allows us to only update the necessary files in a branch, create a PR, then once merged we can use the script to tag and push. Or we can go all in and do it in one step


markphelps:
I tested it by doing all the last releases using this script, I released:

1. flipt-client-go
2. flipt-client-go-musl
3. flipt-client-java
4. flipt-client-java-musl
5. flipt-client-ruby
6. flipt-client-dart
7. flipt-client-python

Will do the JS SDKs once we integrate streaming into them as well

PR #469, opened 2024-10-23
chore: retry if streaming fails

markphelps:
add retries for stream connections

### DIFF
--- a/flipt-engine-ffi/src/http.rs
+++ b/flipt-engine-ffi/src/http.rs
@@ -363,7 +363,12 @@ impl HTTPFetcher {
 
                     while let Some(value) = stream.next().await {
                         match sender.send(value).await {
-                            Ok(_) => continue,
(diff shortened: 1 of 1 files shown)


flipt-engine-ffi/src/http.rs
```diff
@@ -322,59 +322,71 @@ impl HTTPFetcher {
...
                     }
-                }
 
-                Ok(())
+                    // If we've reached here, streaming completed successfully
+                    return Ok(());
+                }
+                Ok(None) => return Ok(()),
+                Err(e) => {
+                    retry_count += 1;
```
GeorgeMac, line 381: Nice: should we set the retry count and backoff back to initial values in the positive case?

To avoid depleting these when network errors are sporadic and short term.

markphelps: yeah good call

Issue #310, opened 2024-07-22
[FLI-1111] C# SDK


markphelps:
### Language

C#

### FFI Support

- [X] I've validated that the language supports Foreign Function Interface (FFI) bindings

### Additional Context

https://github.com/NordSecurity/uniffi-bindgen-cs

https://jakegoulding.com/rust-ffi-omnibus/string_arguments/

https://www.reddit.com/r/rust/comments/e8zus0/calling_rust_from_c_and_vice_versa/

### Willing to Implement

- [X] I'm willing to help implement the Flipt Client SDK in this language if it's accepted. 🙌

### Willing to Review

- [X] I'm willing to review contributions to the Flipt Client SDK in this language if it's accepted.

<sub>[FLI-1111](https://linear.app/flipt/issue/FLI-1111/c-sdk)</sub>

markphelps:
Will likely require a windows build as well

markphelps:
https://learn.microsoft.com/en-us/dotnet/standard/native-interop/native-library-loading

PR #470, opened 2024-10-24
feat: Csharp client

markphelps:
Fixes: #310

Adds C# client!


flipt-client-csharp/src/FliptClient/NativeMethods.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.IO;
+
+namespace FliptClient
+{
+    internal static class NativeMethods
+    {
+        private static IntPtr _nativeLibraryHandle;
+
+        static NativeMethods()
+        {
```
markphelps, line 12: this constructor is not getting called from some reason

cc @skaction

flipt-client-csharp/src/FliptClient/NativeMethods.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        public static T GetDelegate<T>(string functionName) where T : Delegate
+        {
+            // Access a static member to ensure the static constructor is called
+            var handle = _nativeLibraryHandle;
+
+            if (handle == IntPtr.Zero)
+            {
+                throw new InvalidOperationException("Native library not loaded.");
```
markphelps, line 86: this is always true because the static constructor is not setting `_nativeLibraryHandle` because it doesn't seem to be getting called

flipt-client-csharp/src/FliptClient/Models/Response.cs
```diff
@@ -0,0 +1,107 @@
...
+    public class VariantEvaluationResponse
+    {
+        [JsonPropertyName("flag_key")]
+        public string FlagKey { get; set; }
+
+        [JsonPropertyName("match")]
+        public bool Match { get; set; }
+
+        [JsonPropertyName("segment_keys")]
+        public List<string> SegmentKeys { get; set; }
```
markphelps, line 26: i should prob make this a `string[]` to be consistent with other collections elsewere in this file, as we don't really need it to be a List as its size isn't changed after evaluation

flipt-client-csharp/src/FliptClient/FliptClient.csproj
```diff
@@ -0,0 +1,45 @@
...
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <PackageTags>Flipt, Feature Flags</PackageTags>
+    <RepositoryUrl>https://github.com/flipt-io/flipt-client-sdks/tree/main/flipt-client-csharp</RepositoryUrl>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <None Include="../../README.md" Pack="true" PackagePath="\"/>
+  </ItemGroup>
+
+  <ItemGroup>
```
markphelps, line 22: this bit copies the libraries from where they are put in the build pipeline to where they are expected to be loaded from the SDK in https://github.com/flipt-io/flipt-client-sdks/pull/470/files#diff-e065a1af3bac696db30fc1691b298ebd9880d9cc0f1beee156926f5dfd0f58d7R68

test/main.go
```diff
@@ -335,3 +336,28 @@ func dartTests(ctx context.Context, client *dagger.Client, t *testCase) error {
 
 	return err
 }
+
+// csharpTests runs the csharp integration test suite against a container running Flipt.
+func csharpTests(ctx context.Context, client *dagger.Client, t *testCase) error {
+	platform := "linux/amd64"
+	if architecture == "arm64" {
+		platform = "linux/arm64"
+	}
```
markphelps, line 345: we should prob do this for all the tests so that dagger pulls the right Docker image dependent on the runner platform

GeorgeMac (approved):
It's all C# to me

markphelps:
Package looks good:

```
Archive:  Flipt.Client.0.0.1-rc.1.nupkg
  inflating: _rels/.rels             
  inflating: Flipt.Client.nuspec     
  inflating: lib/net8.0/FliptClient.dll  
  inflating: runtimes/linux-x64/native/libfliptengine.so  
  inflating: runtimes/linux-arm64/native/libfliptengine.so  
  inflating: runtimes/osx-x64/native/libfliptengine.dylib  
  inflating: runtimes/osx-arm64/native/libfliptengine.dylib  
  inflating: README.md               
  inflating: [Content_Types].xml     
  inflating: package/services/metadata/core-properties/5a78eaf747c64550be56dda83a76aeae.psmdcp  
  ```

PR #472, opened 2024-10-28
WIP: Swift client

arussellsaw:


flipt-engine-ffi/src/lib.rs
```diff
@@ -212,7 +212,7 @@ pub unsafe extern "C" fn initialize_engine(
 
     let engine_opts_bytes = CStr::from_ptr(opts).to_bytes();
     let bytes_str_repr = std::str::from_utf8(engine_opts_bytes).unwrap();
-    let engine_opts: EngineOpts = serde_json::from_str(bytes_str_repr).unwrap_or_default();
+    let engine_opts: EngineOpts = serde_json::from_str(bytes_str_repr).unwrap();
```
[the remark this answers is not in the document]
markphelps: i guess we panic above though as there are several unwrap calls above this. so ignore me. I've actually been bitten by this before when doing the C# client, where it uses the defaults instead of failing fast so this is probably better they way you have it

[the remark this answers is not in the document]
arussellsaw: ah yes sorry this was from debugging

.github/workflows/test-swift-sdk.yml
```diff
@@ -0,0 +1,57 @@
...
+        env:
+          FLIPT_STORAGE_TYPE: "local"
+          FLIPT_STORAGE_LOCAL_PATH: "./test/fixtures/testdata"
+        run: flipt&
+
+      - name: Install Rust Targets
+        run: |
+          rustup target add aarch64-apple-ios x86_64-apple-ios aarch64-apple-ios-sim
+
+      - name: Build Engine
```
arussellsaw, line 45: this duplicates some work by recompiling the binaries and packaging the xcframework here, but i figured the small time tradeoff is worth the simpler workflow. and this completes before the rest of the integration tests anyway

flipt-client-swift/Sources/FliptClient/FliptClient.swift
```diff
@@ -0,0 +1,282 @@
+import Foundation
+import FliptEngineFFI
+
+public class FliptClient {
+    private var engine: UnsafeMutableRawPointer?
+    private var namespace: String = "default"
+    private var url: String = ""
+    private var authentication: Authentication?
+    private var ref: String = ""
+    private var updateInterval: Int = 0
```
markphelps, line 10: can we make this 120? (seconds)

markphelps (approved):
one minor request for client options defaults, otherwise looks GREAT! ty

package/ffi/main.go
```diff
@@ -588,6 +591,92 @@ func csharpBuild(ctx context.Context, client *dagger.Client, hostDirectory *dagg
...
+	}
+
+	const tagPrefix = "refs/tags/flipt-client-ios-"
+	if !strings.HasPrefix(tag, tagPrefix) {
+		return fmt.Errorf("tag %q must start with %q", tag, tagPrefix)
+	}
+
+	// because of how Go modules work, we need to create a new repo that contains
+	// only the go client code. This is because the go client code is in a subdirectory
+	// we also need to copy the ext directory into the tag of this new repo so that it can be used
```
GeorgeMac, line 660: Could drop or change the messaging here to be swift appropriate

GeorgeMac (approved):
Looks all gravy to me

PR #497, opened 2024-11-04
chore(deps-dev): bump rubocop to 1.68 in flipt-client-ruby

erka:
The older version depends on rexml which has few issues.


markphelps (approved):
ty!

Issue #558, opened 2024-12-05
After go get go.flipt.io/flipt-client this error occurs


nhunam:
```
# go.flipt.io/flipt-client
../../vendor/go.flipt.io/flipt-client/evaluation.go:12:10: fatal error: 'flipt_engine.h' file not found
#include "flipt_engine.h"
```
I'm use MacOS M3. I don't know how to fix it.

markphelps:
@nhunam what version is in your go.mod?

ohh. i see it may be because you are vendoring.. 

I can see `flipt_engine.h` in the latest release when downloading which is what go modules will do. im not sure how vendoring handles it, it may remove non-go files (such as `.h`) files.

![CleanShot 2024-12-05 at 15 28 53](https://github.com/user-attachments/assets/6805f3c5-8813-4d3e-a13e-81752f37b7e8)

let me do some research. cc @GeorgeMac

markphelps:
it looks like by default `go mod vendor` excludes non-go files per https://seankhliao.com/blog/12021-05-09-go-mod-vendor-non-go-things/

we may be able to hack around it though as also described in the blog post

markphelps:
@nhunam can you try to upgrade to the latest version `v0.13.0-rc.3` and see if that resolves your issue?

```
go get -u -v go get go.flipt.io/flipt-client
```

nhunam:
@markphelps : Thank you so much, after upgrade to the latest version, this issue is resolved.

PR #559, opened 2024-12-05
chore: try embed trick to include non-go files on vendor

markphelps:
try to fix: #558

### DIFF
--- a/flipt-client-browser/package-lock.json
+++ b/flipt-client-browser/package-lock.json
@@ -1118,234 +1118,252 @@
       }
     },
     "node_modules/@rollup/rollup-android-arm-eabi": {
-      "version": "4.27.3",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm-eabi/-/rollup-android-arm-eabi-4.27.3.tgz",
-      "integrity": "sha512-EzxVSkIvCFxUd4Mgm4xR9YXrcp976qVaHnqom/Tgm+vU79k4vV4eYTjmRvGfeoW8m9LVcsAy/lGjcgVegKEhLQ==",
+      "version": "4.27.4",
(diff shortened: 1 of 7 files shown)


nhunam (approved):
LGTM

GeorgeMac (commented):
Nice hack. Do we need to do the same for the other non go files?

seems like we will run into this with the dynamic libs too?

Shame we have to then also add all these objects to any resulting binaries in another location in memory for everyone just to make this work 😮‍💨

erka (changes requested):
this doesn't work :(

markphelps:
> this doesn't work :(
> 
> `ext.go` should be in `evaluation` package with
> 
> ```go
> package evaluation 
> 
> import "embed"
> 
> //go:embed ext/*
> var _ embed.FS
> ```

is this because we need to fake embed the libraries too for vendoring? i hope this doesn't make the filesize double in the archive

erka:
@markphelps I tried the local app with vendor vs shared deps and the binary size was the same. It would be nice to have a deeper knowledge about this situation.

erka (approved):

markphelps:
So this is interesting.. it now packages both libc folders in `/ext` and includes them in the archive:

CleanShot 2024-12-06 at 12 43 57

This might actually allow us to implement what is asked for in https://github.com/flipt-io/flipt-client-sdks/issues/141#issuecomment-2469360866 by using buildtags to for the appropriate arch. It still would mean that the initial `go get` would be like 30mb transfered though, but it would only include the needed lib in the actual build of the consuming binary

markphelps:
also its just plain wrong as windows and darwin shouldn't have musl versions.. something odd is going on with the packaging now

markphelps:
ok so i removed the `ext.go` embed and published a new release and can verify the packaging issue is not related to the `ext.go` change, there is something else wrong with our packaging workflow that I'll need to fix first

Issue #593, opened 2024-12-17
[Bug] (Python) Client SDK fails to recover after flipt server restarts


andi4000:
## Problem
Flag evaluation on client SDK continues to fail, even after the flipt server has been restarted. Server logs show that the client SDK is polling/fetching flags, but the evaluation call doesn't recover.

## Steps to Reproduce
1. Spin up a flipt server like so:
```shell
docker run -p 8080:8080 -v ./flipt-data:/var/opt/flipt docker.flipt.io/flipt/flipt:latest
```

2. Add a flag `my-flag`.

3. Run the following python client
```python
from time import sleep

from flipt_client import ClientOptions, FliptEvaluationClient

client = FliptEvaluationClient(opts=ClientOptions(update_interval=2))

while True:
    try:
        res = client.evaluate_boolean("my-flag", entity_id="1337")
        print(res)
    except Exception as e:
        print(e)
    sleep(1)
```

4. Observe that the python client works as expected, printing the evaluation result every second

5. Stop the flipt server

6. Observe the python client correctly logging the following error
```shell
server error: failed to make request: Request failed after 3 retries
```

7. Restart the flipt server

8. ⚠️  Despite the flipt server showing incoming fetch requests from the python SDK, the python client continues to fail with the same error message ⚠️ 

## Expected Behaviour
The client SDK should successfully recover and resume normal operation once the flipt server is back online

## Nasty Workaround
Wrap `evaluate_boolean` and `evaluate_variant` function, handle the (generic) exception and re-initialize `FliptEvaluationClient`

## Environment
- `flipt-client` version `0.13.0`
- Python 3.12
- MBP M2 Pro, macOS 15.2
- Flipt server:
```
Version: v1.51.1
Commit: 7d46879a4ba113916a3901cec7dbf0189db8fc90
Build Date: 2024-11-05T22:18:52Z
Go Version: go1.23.2
OS/Arch: linux/arm64
```

## Additional Notes
Happy to help if further clarification or testing is needed! 😄

markphelps:
Hey @andi4000 

Thanks for raising this issue. Currently this is 'expected' behavior as we chose to not retry indefinitely on failed requests and chose a max of 3 retries. I realize that those 3 attempts happen in rather quick succession so in the case that the Flipt server dies and then is restarted the 3 attempts have likely already happened.

[...]

PR #594, opened 2024-12-18
fix(ffi): clear stored etag on http req failure

markphelps:
Should fix: #593 

This changes the FFI http code to clear the etag on any error responses from the server. This means that on the next request it should re-request fresh state when successful, which will then clear the stored error from the previous request in the snapshot store


erka (approved):
Interesting way of solving it.

PR #615, opened 2025-01-02
chore: try native android sdk test action

markphelps:


erka:
I couldn't get the emulator running for unknown reason. It was okay locally but I messed it up.

The Flipt URL probably should be set to `http://10.0.2.2:8080` as emulator is a separated box

flipt-engine-ffi/build.sh
```diff
@@ -1,13 +1,46 @@
...
+
+"android")
+  rustup target add x86_64-linux-android aarch64-linux-android
+
+  cargo build -p flipt-engine-ffi --release --target=x86_64-linux-android
+  cargo build -p flipt-engine-ffi --release --target=aarch64-linux-android
+  mkdir -p ../flipt-client-kotlin-android/src/main/cpp/libs/x86_64
+  mkdir -p ../flipt-client-kotlin-android/src/main/cpp/libs/arm64-v8a
+  cp ../target/x86_64-linux-android/release/libfliptengine.a ../flipt-client-kotlin-android/src/main/cpp/libs/x86_64/libfliptengine.a
+  cp ../target/aarch64-linux-android/release/deps/libfliptengine.so ../flipt-client-kotlin-android/src/main/cpp/libs/arm64-v8a/libfliptengine.a
```
markphelps, line 37: fwiw I dont think we need the arm libraries as this test should only run on x86 linux on GitHub actions. 

also why `./target/aarch64-linux-android/release/deps/libfliptengine.so` ?

markphelps:
> I couldn't get the emulator running for unknown reason. It was okay locally but I messed it up.
> 
> The Flipt URL probably should be set to `http://10.0.2.2:8080` as emulator is a separated box

This blog post seems to suggest we can run it and doesn't require any third party actions: https://medium.com/innovies-club/run-android-emulator-out-of-the-box-with-github-actions-b84cba766e62

flipt-engine-ffi/build.sh
erka: gradle wants to have all the assets :( 

I  wish I could use outputs from build engine ffi action as input for this. Don't know if it's possible.

markphelps: ah, im curious about two things:

1. why its in the `deps` subfolder
2. why its a dynamic `.so` ?

markphelps:
going to merge this to see if it only triggers the android workflow if on main

PR #619, opened 2025-01-04
chore: test android workflow

markphelps:
@erka this should now download the pre-built android libs from the `package-ffi` workflow and (hopefully) move them to the correct location in the `flipt-client-android-kotlin` directory.

Now all that should be left is to get the emulator to start 🤞🏻


flipt-client-kotlin-android/src/androidTest/java/io/flipt/client/TestFliptEvaluationClient.kt
```diff
@@ -15,14 +14,16 @@ class TestFliptEvaluationClient {
     fun initAll() {
         val fliptURL = BuildConfig.FLIPT_URL
         val clientToken = BuildConfig.FLIPT_AUTH_TOKEN
-
+        assert("http://10.0.2.2:8080" == fliptURL)
```
markphelps, line 17: so Flipt should be running on localhost now right? since its running in the action. Does the emulator give localhost a different IP?

erka: Android emulator is a vm. Similar to docker there are reserved IPs for accessing something in the network or host https://developer.android.com/studio/run/emulator-networking . I debugged it locally with `adb logcat` and everything looks like with ffi options.

For some reason tests could not access network. It doens't matter that the host is. try.flipt.io could not be reach as well. It must be something with permissions or network security policies.

erka:
Just can't crack it.... I tried to add the `AndroidManifest.xml` with `network_security_config.xml` but it doesn't make any difference for me. The emulator doesn't see the external Flipt instance with `gradlew cAT`. It works in webview or Chrome however... I am stuck.

markphelps:
> Just can't crack it.... I tried to add the `AndroidManifest.xml` with `network_security_config.xml` but it doesn't make any difference for me. The emulator doesn't see the external Flipt instance with `gradlew cAT`. It works in webview or Chrome however... I am stuck.

Dang I tried something similar before I saw your comment.. I just dont get why it wouldnt work in actions. Maybe we could try specifying the DNS server (https://developer.android.com/studio/run/emulator-networking#dns) although seems that might not help since we're trying to connect to a local IP

erka:
@markphelps I tried the random sample for okhttp test and run it along side `TestFliptEvaluationClient ` and it works. I hard-coded the url in the ffi engine and it worked (something wrong with list flags but it's minor).  Something is  wrong with serde of engine ffi options and as result it uses default.

markphelps:
Thank you for digging in @erka ! I will take a look at this today and try to determine why the engine isn't getting the correct options.

erka:
@markphelps there is an issue with `authentication`. In kotlin the auth object is serialized to json like `"authentication":{"type":"client_token","client_token":"..."}`. The `type` is unexpected field and serde fails because it can't match anything in emun.

markphelps:
> @markphelps there is an issue with `authentication`. In kotlin the auth object is serialized to json like `"authentication":{"type":"client_token","client_token":"..."}`. The `type` is unexpected field and serde fails because it can't match anything in enum.

legend. thank you for finding this! should be easy to fix

erka:
@markphelps It looks like we crack it... There is few things which should be probably addressed. Flipt instance has not auth right now, right? Another - this awkward not to see the Android action in PR itself...

markphelps:
Thank you @erka for all the hard work on this!!

> There is few things which should be probably addressed. Flipt instance has not auth right now, right? Another - this awkward not to see the Android action in PR itself.

So yeah we could/should probably configure Flipt with auth here. I will do that.

As for the PR / action, once this is merged it should actually run as part of the PR process since the Package FFI Engine runs on PR and then calls the Test Android SDK workflow once complete. Once this is merged we can make it a required check

erka (approved):

PR #655, opened 2025-01-25
fix: Use dagger to detect the arch and platform in stead of go runtime

atmask:
@markphelps Here are my changes to go into your branch if you are okay with them.


markphelps (approved):
nice! ty @atmask

PR #666, opened 2025-01-28
Use ClientOptions type in provider to allow for authentication

tjjevans1:
Was getting a type error when trying to add authentication to the provider.

NB: First time contributing. Realise there may be room for improvement.


markphelps (approved):
Looks great! Thanks @tjjevans1 ! Mind signing the DCA so we can get this merged and released?

Issue #671, opened 2025-02-02
swift sdk not usable per instructions (without modifying source code)


EXC-BAD-ACCESS:
per instructions:
```
import FliptClient

// Initialize the client
let client = try FliptClient(
    namespace: "default",
    url: "http://localhost:8080",
    authentication: .clientToken("your-token"),
    updateInterval: 120,
    fetchMode: .polling
)

// Evaluate a boolean flag
let boolResult = try client.evaluateBoolean(
    flagKey: "my-flag",
    entityId: "user-123",
    context: ["key": "value"]
)
print("Flag enabled: \(boolResult.enabled)")

// Evaluate a variant flag
let variantResult = try client.evaluateVariant(
    flagKey: "my-variant-flag",
    entityId: "user-123",
    context: ["key": "value"]
)
print("Variant key: \(variantResult.variantKey)")

// Don't forget to close the client when you're done
defer {
    client.close()
}
```

the above code results in an error:
<img width="1520" alt="Image" src="https://github.com/user-attachments/assets/e0eb4906-357a-4bf1-8121-6eb35cc8f45a" />

but choosing the fix results in more than just a documentation error:
<img width="1520" alt="Image" src="https://github.com/user-attachments/assets/0349a6d0-fdc5-4c9a-8a15-83e0349b3008" />

because when importing an external package all fields of a struct that you want accessible need to be explicitly marked `public` and any struct you want to be instantiable outside the package need to have an explicit constructor (literally just to mark it public -- it sucks, i know)...

what you probably want is something like this (which i'll put in a PR...)

```
import Foundation
import FliptEngineFFI

public class FliptClient {
    private var engine: UnsafeMutableRawPointer?
    private var namespace: String = "default"
    private var url: String = ""
    private var authentication: Authentication?
    private var ref: String = ""
    private var updateInterval: Int = 0
    private var fetchMode: FetchMode = .polling

    public init(namespace: String = "default",
         url: String = "",
         authentication: Authentication? = nil,
         ref: String = "",
         updateInterval: Int = 120,
         fetchMode: FetchMode = .polling) throws {
        self.namespace = namespace
        self.url = url
        self.authentication = authentication
        self.ref = ref
        self.updateInterval = updateInterval
        self.fetchMode = fetchMode

        let clientOptions = ClientOptions(
            url: url,
            authentication: authentication,
            updateInterval: updateInterval,
            reference: ref,
            fetchMode: fetchMode
        )

[...]

PR #672, opened 2025-02-02
makes struct fields public and adds public init

EXC-BAD-ACCESS:
changes mentioned in https://github.com/flipt-io/flipt-client-sdks/issues/671

*Note i only updated the fields that i was evaluating, didn't touch 
```
ClientTokenAuthentication
JWTAuthentication
Result
```

happy to update those and resubmit if those should be public too

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -70,6 +70,16 @@ updates:
       # kodiak `merge.automerge_label`
       - "automerge"
 
+  - package-ecosystem: gradle
+    directory: "/flipt-client-kotlin-android"
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 5
+    labels:
+      - "dependencies"
+      # kodiak `merge.automerge_label`
+      - "automerge"
+
   - package-ecosystem: bundler
     directory: "/flipt-client-ruby"
     schedule:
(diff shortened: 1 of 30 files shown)


markphelps:
Thank you @EXC-BAD-ACCESS !! As you can probably tell we are not very well versed in the Swift / iOS world!

We probably just want to make these public as well:

```swift
public struct ClientTokenAuthentication: Codable {
    let token: String
}

public struct JWTAuthentication: Codable {
    let token: String
}

public enum FetchMode: String, Codable {
    case streaming
    case polling
}

public struct ClientOptions<T: Encodable>: Encodable {
    let url: String
    let authentication: T?
    let updateInterval: Int
    let reference: String
    let fetchMode: FetchMode
}
```

I don't think we need to make `Result` fields public as those are not meant to be exposed to the caller

markphelps:
@EXC-BAD-ACCESS sorry for the ping but do you mind following the steps here: https://github.com/flipt-io/flipt-client-sdks/pull/672/checks?check_run_id=36661301305 so we can get this merged/released?

Thank you for your help!! ❤️

markphelps:
@EXC-BAD-ACCESS hmm its still showing there's one commit authored by you that isn't signed:

CleanShot 2025-02-06 at 16 38 20

markphelps (approved):
thank you again @EXC-BAD-ACCESS . i'll get this released in the morning

Issue #669, opened 2025-01-31
Flipt Client reports error and no results instead of resolving evaluation request using local data store


ian-a-anderson:
The docs state that once the data has been loaded on the initial request, it should be cached and then updated based on background polling. When testing this we see the local Flipt client evaluation continuing to succeed as expected if we disconnect the server, but this only lasts up until the next polling window. At that point the local cache seems to clear and then no evaluated flags are returned. As per the discussion below, this appears to be a nuance of how the current code treats errors during polling (reporting/throwing that error on the next subsequent evaluation request).

Ideally the client configuration would allow for these types of transient errors to be ignored. Or, alternatively, the client could report the evaluation results along with some sort of indication of the staleness of the data, or a report of the last polling error that occurred prior to the evaluation? Either way, the goal would be to smooth out situations where the local store could continue to successfully resolve evaluation requests with the data it has on file.

Previous discussion via:

> Actually after looking at the code, you're correct @ian-a-anderson. This should be a new issue I think.
> 
> https://github.com/flipt-io/flipt-client-sdks/blob/c2b2b401076de7a8f07fe827f5cc64d1363a0772/flipt-engine-ffi/src/evaluator.rs#L40-L42
> 
> Here we store any errors and return the error at evaluation time. The problem is really that we cant tell which errors are intermittent vs which ones are persistent so we erred on the side of caution by exposing any errors to the caller. However if the next poll loop to pull state succeeds it will clear the error.
> 
> We could probably come up with a better strategy though, maybe allow you the caller to specify how you want errors handled? Or only store errors from the server on first pull and/or after multiple attempts?
> 
> Wdyt? 
> 
> cc @erka  

 _Originally posted by @markphelps in [#148](https://github.com/flipt-io/flipt-client-sdks/issues/148#issuecomment-2628351603)_

ian-a-anderson:
The above behavior was observed with server v1.54.2 and the latest C# client

ian-a-anderson:
@markphelps I just saw the C# client release and updated to [v0.2.0 via NuGet](https://www.nuget.org/packages/Flipt.Client/). However, when I attempted to use the new ErrorStrategy.Fallback option on the ClientOptions it seems to break evaluation completely. If I leave ErrorStrategy unspecified, or if I set it to ErrorStrategy.Fail, then evaluation works. But if I set ErrorStrategy to ErrorStrategy.Fallback then I get no evaluation at all and the following error:
[...]

PR #673, opened 2025-02-02
feat(engine-ffi): add option to evaluate the flag against the stale cache

erka:
related #669

### DIFF
--- a/flipt-engine-ffi/src/lib.rs
+++ b/flipt-engine-ffi/src/lib.rs
@@ -9,7 +9,7 @@ use fliptevaluation::{
     BatchEvaluationResponse, BooleanEvaluationResponse, EvaluationRequest,
     VariantEvaluationResponse,
 };
-use http::{Authentication, FetchMode, HTTPFetcher, HTTPFetcherBuilder};
+use http::{Authentication, ErrorStrategy, FetchMode, HTTPFetcher, HTTPFetcherBuilder};
 use libc::c_void;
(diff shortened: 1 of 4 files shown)


erka:
@markphelps I have updated client options C# only to gather initial feedback before proceeding with broader changes.

flipt-engine-ffi/src/lib.rs
```diff
@@ -96,6 +97,7 @@ impl Default for EngineOpts {
             update_interval: Some(120),
             reference: None,
             fetch_mode: Some(FetchMode::default()),
+            allow_stale: Some(false),
```
markphelps, line 100: how about instead we introduce a new 'error_strategy' option that will allow us to be more flexible in the future (ie we could add retry support at this level too). Example:

```rust
pub enum ErrorStrategy {
    Report,     // (Default) Store the error like we do now
    Fallback,  // Fallback with the previous Snapshot state
}
```

Fallback would be akin to `allow_stale`

markphelps (approved):
looks great! thank you @erka . @ian-a-anderson would this solve your use case? 

once released you would need to set the `error_strategy` option to `fallback` to effectively ignore errors from upstream

ian-a-anderson:
> looks great! thank you @erka . @ian-a-anderson would this solve your use case?
> 
> once released you would need to set the `error_strategy` option to `fallback` to effectively ignore errors from upstream

Sounds good to me! Thanks for the fast update!!

PR #697, opened 2025-02-06
feat: rename report error strategy to fail

markphelps:
Renames `report` to `fail` for 'fail-fast' as I think report might be a bit confusing as we don't actually 'report' the error as in like report to Sentry or anything. I'm open to other names though

Also add a section in the C# readme explaining this new option. I'd like to release the C# client and then add this new option to the other SDKs as well


erka (approved):
naming is hard

PR #718, opened 2025-02-12
build(ffi-engine): build ubuntu-22.04

erka:


erka:
@markphelps I don't know that is wrong with Swift. It looks the runner is failing in other PR too.

markphelps (approved):
ty!

markphelps:
@erka I found this but its a bunch of stuff I don't understand as its iOS specific: https://forums.developer.apple.com/forums/thread/666335

one option it to simply remove x86_64 support for Darwin and say we only support arm

markphelps:
@erka I updated the setup-flipt action to include the ability to run on Mac x86_64

markphelps:
@erka maybe we should just remove the Mac x86 build for now to unblock?

erka:
> @erka maybe we should just remove the Mac x86 build for now to unblock?

Yeah... It's probably the best option right now  because it's tricky to do it for me

PR #719, opened 2025-02-12
experiment: Static musl build

markphelps:
try building a true static lib for musl as previously we were still dynamically linking

For musl targets, we need use `+crt-static` instead of `-crt-static` to enable static linking

`-crt-static` means disabling static linking which Claude says is still required for Android

The idea is that if we use a static lib built with musl we can still use it in hosts that use glibc.. which would mean we could finally get away from the two different versions of libs for Go/Java/etc (glibc vs musl)

This also switches to use the [rusttls](https://crates.io/crates/rustls) crate and feature for `reqwest` which is a TLS lib implemented in Rust instead of depending on the system openssl which we cant link against for musl if we want static links

## TODO:

- [x] Get rid of the glibc builds all together
- [x] Update the packaging code to switch to `.a` instead of `.so` for linux
- [x] Test!


erka (approved):
it looks great and would be awesome to finish this up

markphelps:
So I updated the tests to build the static lib and put them in the right place like we were doing for the dynamic libs in the various SDKS (in `/test/main.go`).

Now we'll need to determine which languages can call into a static lib via FFI or some other mechanism. Go seems to work after I added one `-l` flag (`-lm` to link the math library). But the others are currently failing, although I havent made any changes to them so probably expected.

### Current Status

✅ Go (required adding `-lm` flag)

❌ Ruby (it seems the FFI gem only works for dynamic libs https://github.com/ffi/ffi/issues/225)
❌ Java  (`java.lang.UnsatisfiedLinkError at TestFliptEvaluationClient.java:25)
❌ Dart (trying to load dynamic lib.. `Invalid argument(s): Failed to load dynamic library '/src/lib/src/ffi/linux_x86_64/libfliptengine.a': /src/lib/src/ffi/linux_x86_64/libfliptengine.a: invalid ELF header`)
❌ Python (`OSError: /src/flipt_client/../ext/linux_x86_64/libfliptengine.a: invalid ELF header`)
❌ CSharp (`/src/tests/FliptClient.Tests/bin/Debug/net8.0/runtimes/linux-x64/native/libfliptengine.a: invalid ELF header`)

markphelps:
## New Approach

Go back to building as a `.so` for Linux but on a musl (alpine) host... which _should_ also work on glibc hosts 🤞🏻

markphelps:
Ok.. so now with that approach here are the results:

```
 === Test Results Summary ===
Successful tests:
✅ react
✅ ruby
✅ python
✅ dart
✅ go
✅ csharp
✅ browser

Failed tests:
❌ java: input: container.from.withWorkdir.withDirectory.withFile.withServiceBinding.withEnvVariable.withEnvVariable.withExec.sync resolve: process "gradle test" did not complete successfully: exit code: 1
```

In the tests we're compiling the .so on a linux-musl (alpine) host and then copying it to the test containers for each language which AFAIK are all debian based.. so this should prove that musl `.so`s can be linked and run properly on glibc hosts.

Now we just gotta fix the java one..

markphelps:
Phew I went down some rabbit holes..

Here's a summary of the key changes made to fix the musl/glibc compatibility issues:

In test/main.go:
- Modified the Rust build container to use Alpine Linux
- Added flags to build a dynamic library with musl libc: -C target-feature=-crt-static
- Added symbolic binding optimizations: -Bsymbolic and -Bsymbolic-functions
- **Copied required musl runtime libraries (ld-musl-*.so.1 and libc.musl-*.so.1) alongside our libfliptengine.so**
- Used architecture-specific paths for ARM64/x86_64 support

In flipt-client-go/evaluation.go:
- Simplified the cgo LDFLAGS to properly handle library and runtime paths

The key insight was that building with musl on Alpine allows the library to work on both musl and glibc systems, but we need to bundle the musl runtime libraries with our .so file. This approach avoids requiring users to install musl packages on their systems while maintaining compatibility across different libc implementations.

So.. now we need to test with the other languages.. but Go works on the various distros:

```
=== Test Results Summary ===

Successful tests:
✅ go 1.23-bullseye
✅ go 1.23-bookworm
✅ go 1.23-alpine
```

I suppose its not a big deal to copy the required musl libs `ld-musl-*.so.1 and libc.musl-*.so.1` into our distributed clients alongside our own `libfliptengine.so` if it means now we only need to build one version of the lib for each OS/Arch instead of both a glibc & musl version? @erka, @GeorgeMac wdyt??

erka:
I don't know about other distros but on Debian 13 `/usr/lib/ld-musl-x86_64.so.1` is a symlink to `/usr/lib/x86_64-linux-musl/libc.so`.

markphelps:
Update!

So it looks like option 1 does work at least on x86 linux (via docker) in my initial testing. I likely broke Darwin and Windows support though because now we are creating only a static lib from Rust and not a dylib, and Darwin and Windows can work with dylibs just fine..

But if this passes CI for Go I'll try to get it working for the other languages 🤞🏻 on Linux and test on Debian and Alpine hosts for both x86_64 and arm64.

Then if we can get the other languages to work again we can go back and fix the differences between Linux and Windows/Darwin

erka:
@markphelps could we keep `rlib`? Without it it's impossible to use the crate outside. The examples are failing because of this too.

markphelps:
> @markphelps could we keep `rlib`? Without it it's impossible to use the crate outside. The examples are failing because of this too. 

Ah yes. I can add that back

markphelps:
Ironically now the only tests that are failing are the Go ones (I fixed the dart one locally but for some reason CI hasnt re-triggered)

markphelps:
Ok so I'm consistently getting segfaults when trying to run the Go tests.

I spun up a Linux (Ubuntu) VM and went through the steps to build the `.so` and try to run the tests using `gdb` to debug the segfault:

```console
# this builds the static lib (.a)
cargo build -p flipt-engine-ffi --release
# next we copy it to the flipt-engine-ffi dir where the build.sh script is which creates our .so using musl-gcc
cp target/release/libfliptengine.a flipt-engine-ffi
cd flipt-engine-ffi
./build.sh
# then we move the .so and .h files to the places that the Go SDK expects them (see the existing CGO directives)
cd ..
mkdir -p flipt-client-go/ext/linux_aarch64
cp flipt-engine-ffi/libfliptengine.so flipt-client-go/ext/linux_aarch64/
cp flipt-engine-ffi/include/flipt_engine.h flipt-client-go/ext/
# finally we build our test with debugging symbols added
go test -c -gcflags="all=-N -l"
# and run the compiled test with gdb
gdb ./flipt-client.test
```

within gdb:

```console
(gdb) run -test.v
Starting program: /home/mark/workspace/flipt-client-sdks/flipt-client-go/flipt-client.test -test.v
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/aarch64-linux-gnu/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x0000fffff7d9b900 in getauxval () from /home/mark/workspace/flipt-client-sdks/flipt-client-go/ext/linux_aarch64/libfliptengine.so
(gdb) bt
#0  0x0000fffff7d9b900 in getauxval () from /home/mark/workspace/flipt-client-sdks/flipt-client-go/ext/linux_aarch64/libfliptengine.so
#1  0x0000fffff79efdec in init_have_lse_atomics () at ./lib/builtins/cpu_model/aarch64/lse_atomics/sysauxv.inc:4
#2  0x0000fffff7fc7624 in call_init (env=0xfffffffff080, argv=0xfffffffff068, argc=2, l=<optimized out>) at ./elf/dl-init.c:70
#3  call_init (l=<optimized out>, argc=2, argv=0xfffffffff068, env=0xfffffffff080) at ./elf/dl-init.c:26
#4  0x0000fffff7fc772c in _dl_init (main_map=0xfffff7fff370, argc=2, argv=0xfffffffff068, env=0xfffffffff080) at ./elf/dl-init.c:117
#5  0x0000fffff7fd9cc8 in _dl_start_user () from /lib/ld-linux-aarch64.so.1
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

so we can see the crash is coming from `getauxval()` call

I found this Go issue from 2017: https://github.com/golang/go/issues/13492

Currently reading through it to see if there's any solution.. but I wanted to post my steps to reproduce here incase anyone else had any idea / could reproduce

cc @erka

markphelps:
After doing some more digging I don't think it's gonna work for Go.. seems that segfault is a known issue after reading the above linked GitHub issue and won't be fixed any time soon. 

Also seems that cgo doesn't officially support any c std lib other than glibc..

Anyways I think we'll keep this as is for the other languages that use FFI as it seems to work quite well. 

I think next steps for go cross compatibility is to compile the evaluator to wasm and use something like wazero to call it from Go

PR #724, opened 2025-02-14
chore: split up package-ffi workflows by OS

markphelps:
splits up package ffi workflows by OS to make them more manageable instead of all the `if` checks

## TODO

- [x] Add back notify job in another workflow when all these finish


erka (approved):
nice

markphelps:
I rebased this onto main instead of on the static-musl-build branch because regardless if the latter works or not I'd like to get this in

Issue #737, opened 2025-02-19
Expand support for error-strategy


markphelps:
We added support for specifying error strategy (fail or fallback) in the last version of the FFI engine: #697 and #673

We then exposed this to a few of the SDKS:

- Go #698
- Swift #721
- CSharp #697

We should expand support for error strategy for the following FFI SDKs:

- [x] Android
- [x] Ruby (Released ✅ )
- [x] Python (Released ✅ )
- [x] Java (Released ✅ )
- [x] Dart (Released ✅ )
- [x] Node (Released ✅ )
- [x] Browser (Released ✅ )
- [x] Swift

This includes:

- Adding the errorStrategy option in the source language SDK
- Documenting this new field in the SDKs readme

Also once complete we should likely add a section in our public docs as well (docs.flipt.io)

PR #743, opened 2025-02-21
feat(android): added error strategy option

erka:
related #737


markphelps (approved):
looks great! thank you @erka !!

PR #744, opened 2025-02-24
feat: Go wasm

markphelps:
This PR redoes the Go SDK to use pure WASM to get around the issues with Go and musl v glibc issues as mentioned in #719 and other issues/PRs

## High Level Changes

1. Renamed `flipt-engine-wasm` to `flipt-engine-wasm-js` as it is JS specific since it uses wasm-bindgen to generate JS bindings
2. Created `flipt-engine-wasm` which compiles to 'pure' wasm using the `wasm32-wasip1` target
3. Leverage wazero to handle the  calls to / from wasm and Go **without** requiring CGo
4. Implement the fetching of flag state in pure Go as WASM cant do that yet, similar to how we do in the JS/browser SDKs
5. Changed some public facing API signatures for the Go SDK (see below as we need to decide if we want to do this now or not)

## Go SDK API changes

⚠️  = breaking change

1. ⚠️ I changed `WithUpdateInterval` signature to use a `time.Duration` instead of `int` as its more idiomatic
2. I added the missing `WithRef` option
3. ⚠️ I changed the `New` func to take in a `context.Context` as wazero requires one, although we could undo this change and just use context.Background.. however using a context here will allow the user to set a timeout on the initial state fetch if they wish  
4. ⚠️ I changed `Close` to also take in a `context.Context` for shutdown

## TODO

- [x] implement `FetchMode`, specifically support `Streaming` 😬 
- [x] Implement `ErrorStrategy` which means we'll need to capture the last error from fetch and hold it in the Go code
- [x] Update READMEs / docs
- [x] Benchmark against existing Go FFI based SDK
- [x] pprof to check for memory leaks
- [x] Decide what we want to do about potential breaking changes above. Either undo them or accept them and make this a new major version
- [ ] Probably need to implement retries for fetching like we do currently in the FFI engine
- [ ] Redo packaging for Go SDK (likely in another PR)


markphelps:
I'm going to open this up for review as its already large enough. Will tackle the remaining issues in other PRs before release

markphelps:
cc @jalaziz @piclemx this should finally fix the issues with have multiple versions of the Go SDK, allowing us to collapse it down to one that should work cross-platform

markphelps:
@erka  @GeorgeMac If you get a chance I would love some 👀 on [evaluation.go](https://github.com/flipt-io/flipt-client-sdks/pull/744/files#diff-58b9ecc6b9da5261cb8f7e750248a80ddb464de80acf087070cdcc59f0f47e93) as that is where the meat of the changes are and a bunch of concurrency stuff going on

flipt-engine-wasm/src/lib.rs
```diff
@@ -51,145 +68,286 @@ where
...
+/// # Safety
+///
+/// This function will take in a pointer to the engine and return a batch evaluation response.
+#[no_mangle]
+pub unsafe extern "C" fn evaluate_batch(
+    engine_ptr: *mut c_void,
+    batch_evaluation_request_ptr: *const u8,
+    batch_evaluation_request_len: usize,
+) -> u64 {
+    let e = get_engine(engine_ptr).unwrap();
```
erka, line 223: It would be nice not to panic on the error and just return the error response to the caller.

markphelps: Good shout! This is now handled by the new `result_to_ptr` funcs

markphelps: I also handled this in the FFI lib as well

flipt-client-go/evaluation.go
```diff
@@ -128,151 +124,663 @@ func WithJWTAuthentication(token string) clientOption {
...
+	Result streamResult
+}
+
+type streamResult struct {
+	Namespaces map[string]json.RawMessage
+}
+
+// startStreaming starts the background streaming.
+// Note: currently any errors cause this method to exit. We still need to implement retries for trying to reconnect.
+func (e *EvaluationClient) startStreaming(ctx context.Context) {
```
erka, line 678: Probably there should be some reconnect logic. It will work fine probably to the first connection/timeout error and then client will not get any updates from Flipt (for example, the restart will kill all streaming for the clients).  I don't know if I can emulate this with current open-source version. I just make a wild guess here.

erka (approved):
Overall it looks solid. I have only few comments about it.

flipt-engine-wasm/src/lib.rs
erka: very nice!

flipt-client-go/evaluation.go
```diff
@@ -128,151 +124,663 @@ func WithJWTAuthentication(token string) clientOption {
...
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return snapshot{}, fetchError(fmt.Errorf("unexpected status code: %d", resp.StatusCode))
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return snapshot{}, fetchError(fmt.Errorf("failed to read response body: %w", err))
+	}
```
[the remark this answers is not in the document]
markphelps: good catch! i looked it up and I think you are correct and Claude agrees

flipt-client-go/evaluation.go
markphelps: yeah you are right again! i was thinking of doing the retry/reconnect logic in another PR before release as this one is already quite large. I have it as a todo on the PR  description though.

markphelps:
@jalaziz we just released `flipt-client-go` `v0.16.0-rc.2` which now has no dependency on CGO so it should work the same in both Alpine and GLibc envs

Mind giving it a try when you have a chance and letting us know if it works for you and your team?

Thank you!! 🫶

jalaziz:
> @jalaziz we just released `flipt-client-go` `v0.16.0-rc.2` which now has no dependency on CGO so it should work the same in both Alpine and GLibc envs
> 
> Mind giving it a try when you have a chance and letting us know if it works for you and your team?
> 
> Thank you!! 🫶

@markphelps Just replaced the server-side SDK with the latest version of the client-side SDK. Our unit tests and integration tests are passing, so things are looking great so far!

PR #753, opened 2025-02-28
chore: Packaging for 'static' libs and Go (wasm)

markphelps:
Redoes the packaging code to work with the changes in #747 

I tested this by adding output to stdout using the `tree` command and ensured that all the contents for each SDK are in the correct places

### DIFF
--- a/package/wasm/sdks/go.go
+++ b/package/wasm/sdks/go.go
@@ -16,7 +16,10 @@ type GoSDK struct {
 }
 
 func (s *GoSDK) Build(ctx context.Context, client *dagger.Client, hostDirectory *dagger.Directory, opts BuildOpts) error {
-	const target = "wasm32-wasip1"
+	const (
+		target   = "wasm32-wasip1"
+		wasmFile = "/src/target/wasm32-wasip1/release/flipt_engine_wasm.wasm"
+	)
 
 	rust := client.Container().From("rust:1.83.0-bookworm").
 		WithWorkdir("/src").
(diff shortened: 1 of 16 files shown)


package/wasm/sdks/go.go
```diff
@@ -16,8 +16,29 @@ type GoSDK struct {
...
+		WithWorkdir("/src").
+		WithDirectory("/src/flipt-engine-ffi", hostDirectory.Directory("flipt-engine-ffi")).
+		WithDirectory("/src/flipt-engine-wasm", hostDirectory.Directory("flipt-engine-wasm"), dagger.ContainerWithDirectoryOpts{
+			Exclude: []string{"pkg/", ".gitignore"},
+		}).
+		WithDirectory("/src/flipt-engine-wasm-js", hostDirectory.Directory("flipt-engine-wasm-js"), dagger.ContainerWithDirectoryOpts{
+			Exclude: []string{"pkg/", ".gitignore"},
+		}).
+		WithDirectory("/src/flipt-evaluation", hostDirectory.Directory("flipt-evaluation")).
+		WithFile("/src/Cargo.toml", hostDirectory.File("Cargo.toml")).
```
erka, line 34: should this be a `baseContainer` method in BaseSDK?

markphelps: I think I'll do that if there is another (non-JS) wasm based SDK

erka:
@markphelps could you please double check the workflow to package go sdk? It seems to me that it still uses ffi workflow.

markphelps:
> @markphelps could you please double check the workflow to package go sdk? It seems to me that it still uses ffi workflow.

Gah you are right. Thank you! Fixing now

PR #789, opened 2025-03-04
chore: rm uuid dependency

markphelps:
we dont really need uuids, we just need a string that is unique per flag/rule

this removes the uuid crate which doesn't really work on 'wasm32-unknown-unknown` anyways.. not that it really matters since we are building to `wasip1`.. but its always nice to remove dependencies


erka (approved):
great catch!

PR #790, opened 2025-03-04
fix(go): improve wasm memory allocations and deallocations 

erka:


markphelps (approved):
Looks great thank you 🙏🏻

Issue #755, opened 2025-03-02
Add back connection retries to Go SDK


markphelps:
Since we switched to using WASM for the Go SDK in #744 , we lost the retry mechanism that we do in the FFI based libs (in Rust) to handle connection drops (for streaming) or intermittent connectivity issues for polling

We'll need to handle this natively in Go code since we do the fetching there now

I have a branch [go-retry](https://github.com/flipt-io/flipt-client-sdks/tree/go-retry) that I started that needs to be finished/tested

PR #801, opened 2025-03-10
feat: Go retries

markphelps:
Fixes: #755 

- Adds retries for fetch and initiating streaming (currently 3x) with jitter
- Changes to build the URLs in the constructor, instead of each fetch call

### DIFF
--- a/flipt-client-go/evaluation.go
+++ b/flipt-client-go/evaluation.go
@@ -633,9 +633,14 @@ const (
 	maxDelay   = 2 * time.Second
 )
 
-// fetch fetches the snapshot for the given namespace.
-func (e *EvaluationClient) fetch(ctx context.Context, etag string) (snapshot, error) {
-	var lastErr error
+// doWithRetry executes the given function with exponential backoff and jitter.
+// It will retry up to maxRetries times if the function returns a transient error.
+func doWithRetry[T any](ctx context.Context, fn func() (T, error)) (T, error) {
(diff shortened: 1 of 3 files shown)


flipt-client-go/evaluation.go
```diff
@@ -678,92 +744,110 @@ type streamResult struct {
...
-		e.errChan <- fmt.Errorf("failed to create request: %w", err)
+func (c *EvaluationClient) startStreaming(ctx context.Context) {
+	if err := c.initiateStream(ctx); err != nil {
+		c.errChan <- err
 		return
 	}
+}
+
+func (c *EvaluationClient) initiateStream(ctx context.Context) error {
+	_, err := doWithRetry(ctx, func() (struct{}, error) {
```
erka, line 755: Probably `struct{}` brings no value. Should we remove it?

erka (commented):
I think `startStreaming` should  be like `startPolling`.  It should until the context is cancelled. Currently it looks like any 3 errors will break the streaming. 

There is another thing which I spotted. The `EvaluationClient.err` is never set to `nil` on successful snapshot update and probably will be face the similar bug which was reported for ffi engine. Also simultaneous updates to `err` from `errChan` and `snapshotChan` create potential race conditions and make the final err value unreliable, especially with varying goroutine speeds.

erka (approved):
lgtm

Issue #738, opened 2025-02-19
Support HTTP timeouts


markphelps:
From Discord: https://discord.com/channels/960634591000014878/1341502509558136957/1341502509558136957

Question:

```
We had a question regarding the behavior of our application if Flipt Cloud is unavailable or slow. Specifically:

If Flipt Cloud is down or responding slowly, will our app hang while waiting for a connection/response?
I didn’t see a configurable timeout on the client side (though I may have missed it). Is there a default timeout for connections or responses?
Do you have any recommendations for handling timeouts or failures gracefully?
```

Response:

```
Hey 👋 we dont currently have a configurable timeout for the client sdks but could pretty easily add that.

We also just added support for specifying an error-strategy for the Go and C# SDKs (and can quickly add that to the other ones as well)

The default strategy is fail which means the client will poll Flipt Cloud and retry a few times (3) on error, and if it still doesnt succeed it will return the error to the caller.

The new strategy we added is fallback which will not return an error to the client but will use the 'last known good' state that it got from Flipt Cloud for evaluations.

Then the next successful time the poll loop succeeds, it will replace the old state with the new one
I think we could make this work with timeout too. So if you set a timeout, and that timeout was reached it would use whatever error strategy you set in the client options
```

## Plan

- Implement the ability for users to specify http timeouts for requesting state from Flipt server (or Cloud) in SDKs
- Wire this through to the engine for FFI SDKs
- Wasm based SDKs will need to support this in native code

PR #825, opened 2025-03-18
feat: allow passing http timeout to rust engine for polling

markphelps:
re: #738 

Adds a `request_timeout` option to the http client builder and lib ffi engine opts so that the wrapping SDKs can provide an [overall timeout](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.timeout) for the entire request (only for polling mode).

This should match the `request_timeout` option we have in the [Go SDK](https://github.com/flipt-io/flipt-client-sdks/blob/6452a250f90d2decc68869863e6f6b88503c1266/flipt-client-go/evaluation.go#L139)

Also sets the user-agent header in FFI engine requests to the current engine version (ex: `flipt-engine-ffi/0.6.0`)

And adds a new `Internal` error to the evaluation lib

## TODO (other PRs)

- [ ] add request_timeout option to native SDKs


erka (approved):
nice

PR #828, opened 2025-03-19
chore: More linters

markphelps:
Add more linters for:

- Go
- Kotlin
- C#
- Swift

And fixing linter issues

Note: there are potential breaking changes in the C# SDK, mainly changing some fields to required/nullable as well as some arrays into Lists.. however as we are pre-1.0 there I think it is acceptable in the pursuit of more idiomatic code.


erka (approved):
nice!

PR #855, opened 2025-03-25
fix(go): client streaming url

markphelps:
there were a couple issues with the Go SDK, particularly when used with streaming mode

1. we set the wrong URL for the initial fetch
2. we weren't handling buffered responses like we can get from cloudflare
3. we weren't shutting down correctly, causing the waitgroup in Close to always block

Also made a few improvements:

1. set the user agent header (TODO: automate incrementing of version number in `version.go` on release)
2. don't check engine initialization on each method call, only in `NewEvaluationClient`

### DIFF
--- a/flipt-client-go/evaluation.go
+++ b/flipt-client-go/evaluation.go
@@ -11,6 +11,7 @@ import (
 	"math/rand"
 	"net"
 	"net/http"
+	"strings"
 	"sync"
 	"time"
 
@@ -37,6 +38,10 @@ const (
 	fEvaluateBatch    = "evaluate_batch"
 	fListFlags        = "list_flags"
 	fDestroyEngine    = "destroy_engine"
+
+	// API paths
+	snapshotPath  = "/internal/v1/evaluation/snapshot/namespace/%s"
+	streamingPath = "/internal/v1/evaluation/snapshots"
 )
 
 // EvaluationClient wraps the functionality of evaluating Flipt feature flags.
@@ -49,7 +54,7 @@ type EvaluationClient struct {
 	namespace string
 	err       error
 
-	url            string
+	baseURL        string
 	authentication any
 	ref            string
 	updateInterval time.Duration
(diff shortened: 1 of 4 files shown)


flipt-client-go/evaluation.go
```diff
@@ -228,35 +234,34 @@ func NewEvaluationClient(ctx context.Context, opts ...ClientOption) (_ *Evaluati
...
-	default:
+	// Validate fetch mode
+	if c.fetchMode != FetchModePolling && c.fetchMode != FetchModeStreaming {
 		return nil, fmt.Errorf("invalid fetch mode: %s", c.fetchMode)
 	}
 
+	// Get initial snapshot URL (always uses polling endpoint)
+	initialURL := fmt.Sprintf(c.baseURL+snapshotPath, c.namespace)
+	if c.ref != "" {
+		initialURL += "?reference=" + c.ref
```
erka, line 262: this will fail for Flipt Cloud

markphelps: how so?

erka: The streaming base url has `?` . So adding extra one will break the url

markphelps: its only applied to `snapshotPath` which is `snapshotPath  = "/internal/v1/evaluation/snapshot/namespace/%s"`

erka: Oh... I missed that.

erka:
While I am looking into this PR, I've noticed that engine ffi works a bit different. Previously go client `NewEvaluationClient` never returns error  if initial fetch fails. The error was saved and reported back during the evaluation calls. It fails hard right now  with wasm.

markphelps:
> While I am looking into this PR, I've noticed that engine ffi works a bit different. Previously go client `NewEvaluationClient` never returns error if initial fetch fails. The error was saved and reported back during the evaluation calls. It fails hard right now with wasm.

good shout. maybe we should document this? this is similar to how the other wasm-js SDKs behave now too (node, browser, etc).

We could refactor to make it so the wasm SDKs (including this one) behave more like the ffi ones, where the engine doesn't require a snapshot at initialization time but it would be a non-trivial bit of work, so im not sure how 'worth it' it is

erka:
@markphelps please check #857 for my suggestions for this PR.

erka (approved):
nice

PR #894, opened 2025-04-01
chore: add unknown operator test

markphelps:
add test for cases like in https://github.com/flipt-io/flipt/pull/4018


erka (approved):
nice

PR #898, opened 2025-04-07
chore: Streaming updates ffi

markphelps:
add similar settings for reqwest client like we added in #895

## TODO

- [x] test


markphelps:
I tested with the Ruby SDK by building the `flipt-engine-ffi` lib and copying to the proper place in the ruby codebase. I streamed for several minutes with no issues

Issue #931, opened 2025-04-16
@flipt-io/flipt-client-js does not work in Cloudflare Workers


alexjball:
Cloudflare Workers do not support `WebAssembly.compile`, which is used to load the engine from the inlined wasm module in the browser build. Attempting to initialize the client in a worker env produces this error:

```
[wrangler:err] CompileError: WebAssembly.compile(): Wasm code generation disallowed by embedder
    at _instantiateOrCompile (file:///home/alex/dev/flipt-repro/.wrangler/tmp/dev-mmB1O4/index.js:552:14)
    at _loadWasmModule (file:///home/alex/dev/flipt-repro/.wrangler/tmp/dev-mmB1O4/index.js:569:12)
    at wasm (file:///home/alex/dev/flipt-repro/.wrangler/tmp/dev-mmB1O4/index.js:574:10)
    at FliptClient.init (file:///home/alex/dev/flipt-repro/node_modules/.pnpm/@flipt-io+flipt-client-js@0.0.2/node_modules/@flipt-io/flipt-client-js/src/browser/index.ts:71:22)
    at Object.fetch (file:///home/alex/dev/flipt-repro/src/index.ts:18:35)
    at fetchDispatcher (file:///home/alex/dev/flipt-repro/.wrangler/tmp/bundle-RerTVm/middleware-loader.entry.ts:54:17)
    at __facade_invokeChain__ (file:///home/alex/dev/flipt-repro/node_modules/.pnpm/wrangler@4.11.1_@cloudflare+workers-types@4.20250416.0/node_modules/wrangler/templates/middleware/common.ts:53:9)
    at Object.next (file:///home/alex/dev/flipt-repro/node_modules/.pnpm/wrangler@4.11.1_@cloudflare+workers-types@4.20250416.0/node_modules/wrangler/templates/middleware/common.ts:50:11)
    at jsonError (file:///home/alex/dev/flipt-repro/node_modules/.pnpm/wrangler@4.11.1_@cloudflare+workers-types@4.20250416.0/node_modules/wrangler/templates/middleware/middleware-miniflare3-json-error.ts:22:30)
    at __facade_invokeChain__ (file:///home/alex/dev/flipt-repro/node_modules/.pnpm/wrangler@4.11.1_@cloudflare+workers-types@4.20250416.0/node_modules/wrangler/templates/middleware/common.ts:53:9)
```

Minimal repro here: https://github.com/alexjball/repro-flipt.git

```
pnpm install
[...]

PR #932, opened 2025-04-18
feat: flipt-client-js slim version for edge

markphelps:
re: #931

I tested this in an example cloudflare worker project with the following:

```typescript
import { FliptClient } from "@flipt-io/flipt-client-js/slim";
// @ts-ignore
import wasm from "@flipt-io/flipt-client-js/engine.wasm";

export default {
	async fetch(request, env, ctx): Promise<Response> {
		const client = await FliptClient.init({
			url: '{Flipt Cloud URL}',
			authentication: {
			  clientToken: '{Flipt Cloud API Token}'
			}
		}, { wasm });
		const result = await client.evaluateVariant({
			flagKey: "test",
			entityId: "123",
			context: {},
		});
		return new Response(JSON.stringify(result));
	},
} satisfies ExportedHandler<Env>;
```

and deployed to the local cloudflare worker dev env via `npm run dev`. then viewed the output in the browser env provided by cloudflare devtools:

CleanShot 2025-04-21 at 16 37 35@2x
CleanShot 2025-04-21 at 16 36 42@2x

@alexjball how does this look?


flipt-client-js/src/slim/index.ts
```diff
@@ -0,0 +1,172 @@
...
+      } else if ('jwtToken' in options.authentication) {
+        headers['Authorization'] = `JWT ${options.authentication.jwtToken}`;
+      }
+    }
+
+    let fetcher = options.fetcher;
+
+    if (!fetcher) {
+      // Check if we're in a Node.js environment
+      const isNode = typeof window === 'undefined';
```
erka, line 78: I believe there is no reason in this check because of ` private updateInterval?: NodeJS.Timeout;` and `node-fetch`. Probably it's impossible to run this in browser.

PR #946, opened 2025-04-23
chore: FlagType enum for Python SDK

tkasuz:
This update introduces a `FlagType` enum to represent the `type` field of the `Flag` class in the Python SDK. It enhances clarity and allows easier validation of key types returned by the `list_flags` API.

### DIFF
--- a/flipt-client-python/flipt_client/models.py
+++ b/flipt-client-python/flipt_client/models.py
@@ -1,6 +1,7 @@
 from enum import Enum
-from pydantic import BaseModel, RootModel
 from typing import List, Optional
+
+from pydantic import BaseModel, RootModel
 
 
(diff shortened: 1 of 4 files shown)


markphelps (commented):
nice! thank you @tkasuz !

while you're in here, would you mind adding flag description as well?

https://github.com/flipt-io/flipt-client-sdks/blob/389da4abdcb4c826712de1d658fb1d496bf40de7/flipt-client-js/src/core/types.ts#L177C3-L180C24

Ive been meaning to audit the different sdks and create an issue on which still need these two fields in the flag model

markphelps:
thanks again @tkasuz . I just published the updated [Python SDK v0.19.0](https://pypi.org/project/flipt-client/) with these changes

PR #967, opened 2025-04-30
chore: strip the ios .a libs before dart publish

markphelps:
This is to get around this error on dart publish: https://github.com/flipt-io/flipt-client-sdks/actions/runs/14742115179/job/41382300626#step:7:3573

This removed like 12MB from the uncompressed package which should hopefully let us get under the 100MB limit

CleanShot 2025-04-29 at 20 00 46@2x

This strips the .a files for iOS before adding them to the dart library. I put some TODOs in there as this is a temporary fix as we should:

1. strip them in the package-ios action before adding to the release
2. see if we can produce smaller .a files from the rust lib builds

### DIFF
--- a/package/ffi/sdks/dart.go
+++ b/package/ffi/sdks/dart.go
@@ -109,6 +109,19 @@ func (s *DartSDK) Build(ctx context.Context, client *dagger.Client, container *d
 		}
 	}
 
+	// run strip_xcframework.sh
+	// this is because the XCFramework is not stripped and is too large to upload to pub.dev
+	// TODO: we should do this when we build the XCFramework in Actions but for now this is a quick fix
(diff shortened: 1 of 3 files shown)


package/ffi/sdks/dart.go
```diff
@@ -109,6 +109,19 @@ func (s *DartSDK) Build(ctx context.Context, client *dagger.Client, container *d
 		}
 	}
 
+	// run strip_xcframework.sh
+	// this is because the XCFramework is not stripped and is too large to upload to pub.dev
+	// TODO: we should do this when we build the XCFramework in Actions but for now this is a quick fix
```
markphelps, line 114: now that I added this to the action, we should remove this code before creating the next engine release

PR #971, opened 2025-05-02
chore: try to build smaller libs

markphelps:
experiment to build smaller libs for FFI

https://github.com/johnthagen/min-sized-rust?tab=readme-ov-file#remove-location-details
https://github.com/johnthagen/min-sized-rust?tab=readme-ov-file#remove-fmtdebug

seems to remove about 4-6mb per library which is nice

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ members = ["./flipt-engine-ffi", "./flipt-engine-wasm", "./flipt-engine-wasm-js"
 [profile.release]
 strip = true
 rpath = true
+codegen-units = 1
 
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -13,4 +14,4 @@ opt-level = "s"
 opt-level = "s"
 
 [profile.release.package.flipt-engine-ffi]
-opt-level = "z"
+opt-level = "s"
```
markphelps, line 17: this actually ended up making the lib smaller than `x`

erka (approved):
nice!

Issue #970, opened 2025-05-02
Support Seeding Client SDKs with Serialized Flag State


markphelps:
## Description

Introduce functionality in Flipt client SDKs to allow initializing with a preloaded flag state. This caters to environments where the Flipt server may be inaccessible at runtime, such as air-gapped systems or during initial application startup.

## Proposed Solution

* Add an option to the client SDKs to accept an initial flag state as a serialized string or byte array.
* Provide a method to retrieve the current in-memory flag state, enabling applications to persist it for future use.

## Use Cases

* Applications deployed in environments without network access to the Flipt server.
* Reducing startup latency by avoiding initial network calls to fetch flag states.
* Ensuring consistent flag states across deployments by using a known serialized state.[

## References

* Community discussion: [Client side evaluation with initial read from file system](https://community.flipt.io/t/client-side-evaluation-with-initial-read-from-file-system/99)
* #945

PR #1000, opened 2025-05-13
feat: add the ability to get and set the snapshot in the FFI engine

markphelps:
Re: #970

Adds set/get_snapshot methods to the FFI engine 

We'll need to expose these to the clients, likely starting with the mobile clients

This also removes some `unwraps` in the FFI engine and applies `clippy` suggestions


markphelps:
> it would nice to have a test for `set_snapshot`

we kind of already do as it just calls `replace_snapshot` on the evaluator which has tests

markphelps:
@cristianonicolai this is taking a bit longer than expected, but please leave any comments on #1024 if I'm doing something non-idiomatic java-wise

PR #1024, opened 2025-05-29
feat(java): expose snapshot api

markphelps:
Re: #970

Re: https://github.com/flipt-io/flipt-client-sdks/pull/1000#issuecomment-2912333926


flipt-client-java/src/main/java/io/flipt/client/FliptEvaluationClient.java
```diff
@@ -348,6 +352,31 @@ public ArrayList<Flag> listFlags() throws EvaluationException {
     return resp.getResult().get();
   }
 
+  /**
+   * setSnapshot sets the snapshot for the Flipt client.
+   *
+   * @param snapshot the snapshot to set
+   */
+  public Result<String> setSnapshot(String snapshot) throws EvaluationException {
```
markphelps, line 360: returning the result was just for debugging, will remove this and make it throw an exception if result is not successful

flipt-evaluation/src/models/source.rs
```diff
@@ -46,7 +49,8 @@ pub struct Flag {
 pub struct Variant {
     pub id: String,
     pub key: String,
-    pub attachment: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub attachment: Option<String>,
```
markphelps, line 53: now that we are 'exposing' this data publicly we need to be careful about optionals as this was previously writing the string "null" like {"attachment": "null"} which is not what we want

flipt-evaluation/src/store.rs
```diff
@@ -46,7 +46,7 @@ impl Snapshot {
     pub fn empty(namespace: &str) -> Snapshot {
         Self {
             namespace: Namespace {
-                _key: namespace.to_string(),
+                key: namespace.to_string(),
```
markphelps, line 49: i dont remember why this was `_key` in the first place. do you @erka ?

markphelps:
> @markphelps quick question around this, will it be available/compatible with v1 and v2 servers?

@cristianonicolai yes it will! i just have to do some work to support environments for v2 in the client SDKs: https://docs.flipt.io/v2/concepts#environments

markphelps:
@cristianonicolai I just released `flipt-client-java-v0.17.0-rc.1` which has the new methods for getting/setting the snapshot state per https://github.com/flipt-io/flipt-client-sdks/tree/main/flipt-client-java#snapshotting

would you mind giving it a try and let me know if it fits your usecase?

cristianonicolai:
@markphelps, from what I understood, we need to make the client connect to a running Flipt server in order to get the snapshot before storing it correct? Since we plan to have this as a resource of the app, so it can be used in environments where Flipt server is not available, would it make sense to have something that could generate that snapshot and store it, perhaps a Github Action? The workflow would be to keep a state snapshot file, that would be re-generated every time we change Flipt git configuration. Wdyt?

markphelps:
> @markphelps, from what I understood, we need to make the client connect to a running Flipt server in order to get the snapshot before storing it correct? Since we plan to have this as a resource of the app, so it can be used in environments where Flipt server is not available, would it make sense to have something that could generate that snapshot and store it, perhaps a Github Action? The workflow would be to keep a state snapshot file, that would be re-generated every time we change Flipt git configuration. Wdyt?

@cristianonicolai 

> from what I understood, we need to make the client connect to a running Flipt server in order to get the snapshot before storing it correct

that's correct. the thought was you would be able to connect to flipt server at least once to get the snapshot from the clients and then save the state elsewhere to use as a seed for other clients, but i can see the reasoning for wanting to create the state without having the clients in the loop.

> would it make sense to have something that could generate that snapshot and store it, perhaps a Github Action? The workflow would be to keep a state snapshot file, that would be re-generated every time we change Flipt git configuration. Wdyt?

yeah we could add a snapshot command to the Flipt binary and then support it in our existing GH Action (https://docs.flipt.io/tooling/github-actions) similar to how we already have a `bundle` and `export` commands.

How would you imagine the saving/storing of state would work after the GH Action generated the snapshot file? Would this be something you would manually download from the action output and add as a resource in your app?

markphelps:
so I just realized that the snapshot that gets served from the Flipt server and the one that the client SDKs support now since the 1.0.0 releases are slightly different formats.

here is the code that generates the JSON for the server: https://github.com/flipt-io/flipt/blob/main/internal/server/evaluation/data/server.go which maps to [source::Document](https://github.com/flipt-io/flipt-client-sdks/blob/main/flipt-evaluation/src/models/source.rs) on the client side

the client sdks take the source::Document payload and convert it to a Snapshot [here](https://github.com/flipt-io/flipt-client-sdks/blob/main/flipt-evaluation/src/models/snapshot.rs#L42)

So i think there are 3 options to make what we discussed above work:

1. update the Flipt server side to generate the snapshot JSON as expected in the client [here](https://github.com/flipt-io/flipt-client-sdks/blob/main/flipt-evaluation/src/models/snapshot.rs). 
   
To make this backward compatible we could use request headers so the client can say it wants the data in this format instead of the 'Document' format. Then we wouldnt need to update the client sdks other than the request headers

2. keep the document datatype returned from the server but update the client side snapshot parsing to be able to work with either the current format or the 'raw' Document. we could likely depend on the fact that the snapshot format is [versioned](https://github.com/flipt-io/flipt-client-sdks/blob/ac4c5bb5a2e7ed1c5e188d63ae772ad644618565/flipt-evaluation/src/models/snapshot.rs#L191-L199) so we can quickly detect which format we have 

Again we'd need to be careful not to break backward compatibility

3. just have this new snapshot command in Flipt generate the expected snapshot format, which would mean porting over the logic in the [build](https://github.com/flipt-io/flipt-client-sdks/blob/main/flipt-evaluation/src/models/snapshot.rs#L42-L201) function client side to Flipt in Go. which likely wont be that hard and we can write tests to verify correctness.


ideally I would like to get rid of the [source::Document](https://github.com/flipt-io/flipt-client-sdks/blob/main/flipt-evaluation/src/models/source.rs) format all together so we can stop doing the conversion in the clients.. but doing 3 now doesnt prevent us from doing 1 in the future


wdyt @erka as I know we've briefly discussed the different formats before

cristianonicolai:
@markphelps, just checking if this is something that is still on your radar to be addressed in the near future. Thanks

PR #1025, opened 2025-05-30
Use shim for useSyncExternalStore

TomEBGN:
Adding this shim allows the Flipt react SDK to be used in react 17.

Shim package [here](https://www.npmjs.com/package/use-sync-external-store).

It only uses the shim if the hook is not present in the version of react that you're using.


markphelps (approved):
lgtm! Thanks @TomEBGN and @JordanPBGN

markphelps:
Released: https://www.npmjs.com/package/@flipt-io/flipt-client-react

PR #1026, opened 2025-06-01
feat(dart): add snapshot set/get

markphelps:
add get/setSnapshot support for dart sdk


erka (approved):
nice!

@markphelps it would be nice to have the test where client couldn't connect to Flipt. Like with wrong url or so.

PR #1027, opened 2025-06-02
feat: Add HTTP Client override option

nmishin-lucky:
Adds the option to override default http client in Go SDK which is useful in cases where custom dialer is required.

### DIFF
--- a/flipt-client-go/evaluation.go
+++ b/flipt-client-go/evaluation.go
@@ -94,6 +94,13 @@ func WithURL(url string) ClientOption {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


markphelps:
Thank you @nmishin-lucky ! Releasing v0.17.0 now with this change

markphelps:
@BrunoSzczuk sorry about that. I kicked it off yesterday but there was a GitHub Token issue in the action. I fixed it and re-released it https://github.com/flipt-io/flipt-client-go/releases/tag/v0.17.0

PR #1055, opened 2025-06-05
feat!(ffi): support environments for v2

markphelps:
1. moves namespace to engine opts
2. adds support for environment in engine opts
3. adds `X-Flipt-Environment` header if env is not "", which will work with v2 and still works backward compat for v1


markphelps:
Note: all the tests are going to fail until we update the clients to use the new opts[namespace]

erka (approved):
it will be some extra work for people to do upgrade

Issue #1059, opened 2025-06-05
v2: Environment Support


markphelps:
Tracking issue to keep track of which clients we need to update to support Flipt v2, specifically [environments](https://docs.flipt.io/v2/concepts#environments)

## TODO

- [x] Java
- [x] Kotlin
- [x] Dart
- [x] Ruby
- [x] CSharp
- [x] Go
- [x] JS
- [ ] React
- [x] Python
- [x] Swift

PR #1071, opened 2025-06-09
feat!(react): add env support and move namespace to options

markphelps:
Closes: #1059 

- adds ability to set env for react client
- moves namespace/env to options


erka (approved):
nice

PR #1106, opened 2025-06-18
Fix Pydantic v1 compatibility in Python SDK

ojii:
The Python SDK is, according to the `pyproject.toml`, compatible with Pydantic V1. There is also some compatibility code in `models.py`, however that is not used consistently. This PR aims to use those compatibility shims everywhere.

Since I could not figure out how to run the tests for this, I instead manually tested it in the project I am attempting to integrate this into, though that testing was very shallow.


markphelps:
I released 1.0.1 of the python sdk with this fix. Thanks again @ojii !

Issue #1107, opened 2025-06-18
JS SDK: errors from the WASM engine lack the reason for the error


dedwardstst:
Language Client: flipt-client-js@0.1.0
Platform: Node v22.15.1 on MacOS Sequoia 
 
Reproduction (at least how I discovered this):
- setup the flipt client
- create an evaluation context that contains something other than strings ex. `{ "isAgent": true }`
- pass the context to any of the evaluate functions
- this should produce an error
- `console.error` should show `Error` + the stack trace without any sort of description

In digging through the WASM code I found, I found that error messages are passed back as `error_message`
https://github.com/flipt-io/flipt-client-sdks/blob/main/flipt-engine-wasm/src/lib.rs#L60

However the types -> https://github.com/flipt-io/flipt-client-sdks/blob/main/flipt-client-js/src/core/types.ts#L31
And relevant error handling
https://github.com/flipt-io/flipt-client-sdks/blob/main/flipt-client-js/src/core/base.ts#L106
https://github.com/flipt-io/flipt-client-sdks/blob/main/flipt-client-js/src/core/base.ts#L152
https://github.com/flipt-io/flipt-client-sdks/blob/main/flipt-client-js/src/core/base.ts#L175
https://github.com/flipt-io/flipt-client-sdks/blob/main/flipt-client-js/src/core/base.ts#L236

All expect the message to come back as `errorMessage`. I confirmed that switching the error handling to instead look for `error_message` resolves this issue.

dedwardstst:
Also want to note that I have a fix ready to be PR'd that changes the type and error handling to look for `error_message` instead of `errorMessage`.

I held off on sending it, because I wasn't sure if it was preferable to modify the WASM engine code instead. I can also PR changes to the WASM engine response if that's preferable.

erka:
> Like this?
> 
> ```
> #[derive(Serialize)]
> struct WASMResponse<T>
> where
>     T: Serialize,
> {
>     status: Status,
>     result: Option<T>,
>     #[serde(rename = "errorMessage")]
>     error_message: Option<String>,
> }
> ```
Yeah... it should be work like that

erka:
@dedwardstst 

I tried building a small app using `flipt-client-js@0.1.0`, and I see the expected error message, like this:

![Image](https://github.com/user-attachments/assets/b0aa20ed-0722-4b67-b849-b3d608c20c23)

Do you have an example of the error you're seeing?

[...]

PR #1108, opened 2025-06-18
fix: added serde to WASMResponse.error_message in flipt-engine-wasm

dedwardstst:
Fixes: #1107 

Currently in flipt-client-js, errors from the wasm-engine lack a reason/description for the error.

The wasm engine is returning a result like this
```
{
    status: ...,
    result: ...,
    error_message: ""
}
```

but the client is expecting
```
{
    status: ...,
    result: ...,
    errorMessage: ""
}
```

This PR renames the error_message property to `errorMessage`.


markphelps (approved):
Ty!

PR #1133, opened 2025-06-27
feat: add TLS configuration support to FFI engine

markphelps:
## Summary

Implements Phase 1 of self-signed certificate support by adding comprehensive TLS configuration to the FFI engine. This enables FFI-based client SDKs (Python, Java, C#, Ruby, Dart) to connect to Flipt servers using self-signed certificates.

### Core Implementation
- **New `TlsConfig` struct** in `EngineOpts` with support for:
  - `ca_cert_file` / `ca_cert_data` - Custom CA certificates (file path or base64 encoded)
  - `insecure_skip_verify` - Skip certificate verification (development only)
  - `client_cert_file/data` + `client_key_file/data` - Client certificates for mutual TLS

### HTTP Client Integration
- Added `tls_config()` method to `HTTPFetcherBuilder`
- Implemented `configure_tls()` function that configures `reqwest::Client` with custom TLS settings
- Proper error handling with descriptive error messages

### Testing
- **13 comprehensive unit tests** covering:
  - All TLS configuration combinations
  - Error handling for invalid certificates/files
  - Both file-based and data-based certificate loading
  - EngineOpts JSON serialization/deserialization

## Usage Example

FFI clients can now pass TLS configuration via JSON:

```json
{
  "url": "https://localhost:8443",
  "tls_config": {
    "ca_cert_file": "/path/to/ca.crt",
    "insecure_skip_verify": false
  }
}
```

Or with base64-encoded certificate data:

```json
{
  "url": "https://localhost:8443", 
  "tls_config": {
    "ca_cert_data": "LS0tLS1CRUdJTi...",
    "client_cert_data": "LS0tLS1CRUdJTi...",
    "client_key_data": "LS0tLS1CRUdJTi..."
  }
}
```

## Security Considerations

- **Secure by default**: All TLS options are optional, existing behavior unchanged
- **Development safety**: `insecure_skip_verify` requires explicit opt-in
- **Clear error messages**: Detailed feedback for certificate configuration issues
- **Proper certificate validation**: Uses reqwest's built-in certificate handling

## Testing

All tests pass including new TLS-specific tests:
- ✅ Custom CA certificate loading (file and data)
- ✅ Client certificate authentication
- ✅ Insecure mode for development
- ✅ Error handling for invalid configurations
- ✅ JSON serialization compatibility

## Next Steps

This implements the foundation for self-signed certificate support. Future phases will:
1. **Phase 2**: Update FFI client SDKs (Python, Java, C#, Ruby, Dart) to expose TLS options
[...]

PR #1134, opened 2025-06-28
Feat: python tls support

markphelps:
builds upon #1133 by implementing tls config in the Python SDK


test/fixtures/tls/generate_certs.sh
erka, line 1: It would be ideal to generate the test certificates dynamically by running this script prior to executing other tasks, rather than committing them to version control. This  could also be implemented in Go and we have in Flipt repo similar code. The fixture certificates are valid for one year, after which the tests will start failing unless they are renewed.

PR #1136, opened 2025-06-29
feat: add TLS configuration support to Ruby SDK

markphelps:
## Summary

- Add comprehensive TLS configuration options to the Ruby client SDK, matching the functionality from Python and Java implementations
- Add TlsConfig class with Ruby-idiomatic patterns (keyword arguments, attr_reader, factory methods)
- Support CA certificates (file paths and raw data), client certificates for mutual TLS, and insecure mode for development
- Update integration test framework to enable HTTPS testing for Ruby SDK with TLS certificates
- Update Ruby tests to handle HTTPS URLs with TLS configuration automatically like Python/Java
- Add comprehensive README documentation with Ruby-specific TLS examples

## Key Features

- **Ruby-idiomatic design**: Uses keyword arguments, attr_reader, and factory methods
- **File validation**: Automatically validates certificate file existence
- **Convenience methods**: `TlsConfig.insecure`, `TlsConfig.with_ca_cert_file`, `TlsConfig.with_mutual_tls`, etc.
- **Data precedence**: Raw certificate data takes precedence over file paths
- **JSON serialization**: Proper `to_h` method for FFI layer integration
- **Test integration**: Updates integration test framework to use HTTPS with TLS certificates

Issue #1132, opened 2025-06-27
Client SDKs: Support self-signed certs


markphelps:
### Discussed in https://github.com/orgs/flipt-io/discussions/4366

<div type='discussions-op-text'>

<sup>Originally posted by **cristianonicolai** June 20, 2025</sup>
I'm following the configuration to setup HTTPS port on the Docker server with _server.cert_file_ and _server.cert_key_, but I was wondering if the Java and React clients require any special configuration to connect via HTTPS? Thanks</div>

### FFI

- [x] Add support in FFI engine (#1133) released ✅ 
- [x] Add to Python SDK (#1134) released ✅ 
- [x] Add to Java SDK (#1135) released ✅ 
- [x] Add to Ruby SDK (#1136) released ✅ 
- [x] Add to CSharp SDK (#1137) released ✅ 
- [x] Add to Dart SDK (#1137) not released ❌  (see #1139)
- [ ] Add to Swift SDK
- [ ] Add to Kotlin/Android SDK

### WASM JS

- [ ] Implement in JS SDKs

### Pure WASM

- [x] Implement in Go SDK (#1138) released ✅

markphelps:
superseded by #1170

PR #1138, opened 2025-06-29
feat: add TLS configuration support to Go SDK

markphelps:
Re: #1132 

Like: #1137 , #1136 , #1135


flipt-client-go/config.go
```diff
@@ -91,6 +92,40 @@ func WithRequestTimeout(timeout time.Duration) Option {
 	}
 }
 
+// WithTLSConfig sets the TLS configuration for the client using the standard library tls.Config.
+// This provides maximum flexibility for configuring TLS settings including custom CAs,
+// mutual TLS authentication, and certificate verification options.
+func WithTLSConfig(tlsConfig *tls.Config) Option {
```
erka, line 98: The order of options can lead to unexpected behavior. For example, using `WithTLSConfig` followed by `WithHTTPClient`  may not work as intended.
While this combination is probably rare in real world, I’d suggest mentioning the order dependency in the Go doc to avoid confusion.

markphelps: good point. will update the godocs and readme to explain this

PR #1314, opened 2025-09-09
fix(streaming): updated streaming endpoint URL

causton81:
I needed this patch to get streaming to work with flipt v2.1.0. 
Also needed https://github.com/flipt-io/flipt/pull/4715

PR #1315, opened 2025-09-10
fix(client-go): update streaming URL and response format for Flipt v2.1.0

markphelps:
Fixes the streaming URL and response type for the Go client SDK to work with Flipt v2.1.0, similar to the fixes made in PR #1314 for the FFI engine.

## Changes

1. **URL Path**: Updated from `/internal/v1/evaluation/snapshots?[]namespaces=%s` to `/client/v2/environments/%s/namespaces/%s/stream`

2. **Response Format**: Simplified `streamChunk` struct to expect `{result: <Document>}` format instead of the nested namespaces map

3. **Stream Processing**: Updated to directly use the `result` field as the snapshot payload

## Testing

Please test with Flipt v2.1.0 to ensure streaming functionality works correctly.

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/flipt-client-go/client.go
+++ b/flipt-client-go/client.go
@@ -716,6 +716,9 @@ func (c *Client) handleStream(ctx context.Context, r io.ReadCloser) error {
 					// Send the result directly as the payload
 					select {
 					case <-ctx.Done():
+						if ctx.Err() == context.Canceled {
+							return nil
+						}
 						return ctx.Err()
 					case c.snapshotChan <- snapshot{payload: chunk.Result}:
 					}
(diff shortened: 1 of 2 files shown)


erka (commented):
It looks good to me but it would be nice to have the test for it too.

markphelps:
Workin on it in #1317

markphelps:
@claude can you help me add a test for streaming in the Go SDK? it should use streaming mode and wait a second before performing the evaluation calls

erka (approved):

PR #1476, opened 2025-12-01
chore: update Android build script to use 16kb page size

lampemw:
Android is requiring developers to update their apps to support 16KB page size in the near future. I've updated the build script for libfliptengine.so to use 16KB page size. 

https://developer.android.com/guide/practices/page-sizes

I've tested this myself by running the workflow and inserting the updated libfliptengine.so into my flutter app before building and can confirm that it meets the android 16KB page size requirements.

### DIFF
--- a/.github/workflows/package-ffi-engine-android.yml
+++ b/.github/workflows/package-ffi-engine-android.yml
@@ -5,6 +5,7 @@ on:
     paths:
       - "flipt-engine-ffi/**"
       - "flipt-evaluation/**"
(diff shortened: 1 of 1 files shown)


erka:
@markphelps could we do the Android release, please?

lampemw:
@markphelps Sure, I appreciate the quick turnaround. I will take a look at splitting the current flipt_client package into platform specific packages, that's a common approach that many other packages use in the dart ecosystem. I will reach out if I need anything.

PR #1533, opened 2026-01-07
fix(csharp): support arch-specific publish layouts for native library loading

### LINKED ISSUE #1532: Arch-specific published code failing to find Flipt native dependency
Flipt.Client version: 1.3.1
Target dotnet framework: net8.0

We are switching our deployment system to publish architecture-specific images to reduce container size. However, we are hitting an issue with Flipt Client that appears to be due to the native client dependency being stored at the root level when published for a specific architecture. 

Specifically the error we are hitting is this:

```
System.TypeInitializationException: The type initializer for 'FliptClient.NativeMethods' threw an exception.
 ---> System.IO.FileNotFoundException: Native library not found. Searched:
/apps/src/Apps/Apps.Admin.Api/bin/Release/net8.0/linux-arm64/publish/linux-arm64/native/libfliptengine.so
/apps/src/Apps/Apps.Admin.Api/bin/Release/net8.0/linux-arm64/publish/runtimes/linux-arm64/native/libfliptengine.so
```

Note that the paths searched include the `runtimes/linux-arm64/native` and an attempt without the parent "runtimes" folder at `linux-arm64/native`. The "runtimes" folder, which houses the collection of libfliptengine native files per architecture, does get created when published without a target architecture. But we are seeing this file get copied to the root level of the output folder (eg: this file is located at `../publish/libfliptengine.so`) with no corresponding runtime or arch subfolders when publishing with a specific architecture via `dotnet publish --arch arm64`

Is this a known issue with arch-specific publishing? It seems like the quickest fix would be for Flipt to add the root output path into the search paths in this scenario. For now we are working around this by manually copying the published native file to the expected location.

erka:
.NET arch-specific publishing may place the native engine at the publish root.
Include root location in the native library search paths.

closes #1532

### DIFF
--- a/flipt-client-csharp/src/FliptClient/NativeMethods.cs
+++ b/flipt-client-csharp/src/FliptClient/NativeMethods.cs
@@ -73,20 +73,25 @@ namespace FliptClient
         {
             string libraryName = GetPlatformLibraryPath();
             string baseDir = AppDomain.CurrentDomain.BaseDirectory;
-            string directPath = Path.Combine(baseDir, libraryName);
-            string runtimePath = Path.Combine(baseDir, "runtimes", libraryName);
 
-            if (File.Exists(directPath))
+            var searchPaths = new[]
             {
-                return directPath;
+                Path.Combine(baseDir, libraryName),
+                Path.Combine(baseDir, "runtimes", libraryName),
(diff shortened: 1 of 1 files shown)


markphelps (approved):
looks great thank you @erka i'll push a new release

PR #1688, opened 2026-04-03
Fix: swallow unhandled rejection in auto-refresh timer

amin7x6:
## Problem

The `setInterval` callback in `setupAutoRefresh` / `startAutoRefresh` is `async` but has no `try/catch`. When `ErrorStrategy.Fail` is set, `refresh()` re-throws on any network error. The re-throw escapes the timer callback with no handler, becoming an unhandled promise rejection, which terminates the process in Node.js v15+.

## Root cause

```ts
  setInterval(async () => {
    await this.refresh(); // throws for ErrorStrategy.Fail — nobody catches it
 }, interval);
```

## Fix

Wrap the timer callback in try/catch. refresh() keeps its re-throw behavior intact for direct callers — we only handle the unhandled case in the background timer where there is no caller:

```
setInterval(async () => {
    try {
      await this.refresh();
    } catch {
      // cached snapshot from last successful fetch remains valid
    }
  }, interval);
```

## Affected files

  - flipt-client-js/src/node/index.ts
  - flipt-client-node/src/index.ts


markphelps (approved):
looks great thank you!

markphelps:
@amin7x6 do you mind signing the [DCO](https://github.com/flipt-io/flipt-client-sdks/pull/1688/checks?check_run_id=71960391986) so we can merge this and create a release? thank you for your contribution!

amin7x6:
> @amin7x6 do you mind signing the [DCO](https://github.com/flipt-io/flipt-client-sdks/pull/1688/checks?check_run_id=71960391986) so we can merge this and create a release? thank you for your contribution!

signed off all commits, all good now :)

markphelps:
Thanks @amin7x6 ! I redid this in #1711 as there were some CI changes I needed to make, but your commit is still accredited for the fix! Will release a new version of the js SDKs now

PR #1731, opened 2026-05-14
fix(android): add max-page-size linker flag for 16KB device compatibility

### LINKED ISSUE #1730: flipt-client-android:1.3.0 still appears to package a non-16 KB-aligned arm64-v8a native library
Im using io.flipt:flipt-client-android:1.3.0 and Android Studio warns that the APK is not compatible with 16 KB page size devices because of Flipt’s packaged native library:

```
APK app-dev-debug.apk is not compatible with 16 KB devices. Some libraries have LOAD segments not aligned at 16 KB boundaries:
lib/arm64-v8a/libfliptengine_wrapper.so

```

Image

I found PR #1476, which looks like it was intended to address this:

https://github.com/flipt-io/flipt-client-sdks/pull/1476

However, inspecting the published 1.3.0 AAR still seems to show a 4 KB-aligned libfliptengine_wrapper.so for arm64-v8a.

erka:
closes #1730


markphelps (approved):
ty! will create a new release

PR #1732, opened 2026-05-17
feat(ffi): reconnect streaming on auth credential rotation

vss96:
## Summary

A streaming request fixes its `Authorization` header at connect time, so a credential rotated via `update_authentication` never reaches an in-flight stream. This adds a reconnect signal so streaming mode applies rotated credentials without restarting the engine.

- `update_authentication` writes the new headers to the existing auth `watch` channel (as before) and, if they changed, fires a `reconnect_notify`.
- The streaming read loop selects on `reconnect_notify` alongside the existing stop signal — the same `Notify` + `select!` idiom #1482 introduced for shutdown. On signal it drops the connection and reconnects; `build_headers()` then supplies the new credentials.
- A no-op rotation (unchanged headers) skips the signal, so a redundant refresh doesn't needlessly drop the connection.

Polling already applied rotated credentials on its next fetch; this closes the streaming gap left by the dynamic-auth work in #1608.

## Scope

Deliberately minimal — only what streaming auth refresh requires (~390 lines, mostly tests). No retry/backoff overhaul, cancellable-connect, or liveness flag; those are independent hardening concerns and out of scope here.

## Test plan
- [x] `cargo test -p flipt-engine-ffi` — 39 passed
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- New tests: streaming reconnects on rotation, skips reconnect when auth unchanged, clears auth via `"none"`, rejects `{}`, and the full FFI rotation path against a live server.

### DIFF
--- a/flipt-engine-ffi/src/http.rs
+++ b/flipt-engine-ffi/src/http.rs
@@ -101,6 +101,10 @@ pub struct HTTPFetcher {
     namespace: String,
     auth_receiver: watch::Receiver<HeaderMap>,
     auth_sender: Option<watch::Sender<HeaderMap>>,
+    /// Signals the streaming loop to drop its current connection and reconnect,
+    /// picking up rotated credentials. A streaming request's headers are fixed
+    /// at connect time, so new credentials only take effect on a new connection.
+    reconnect_notify: Arc<Notify>,
     etag: Option<String>,
     reference: Option<String>,
     update_interval: Duration,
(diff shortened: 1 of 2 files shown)


erka:
Hey @vss96 

Why does the stream need to reconnect? I feel that I'm missing some understanding of the problem.

vss96:
@erka my initial thought for streaming was that the token would remain stale once the connection is open and we don't update the auth like we do for the polling clients.
I see that it's probably okay for a stale token to exist on a stream that is established because you can keep sending updates over the stream to the client, so maybe this is not required.

Issue #1480, opened 2025-12-06
feat: offline support


luisbytes:
Currently, the feature flag system requires network connectivity to retrieve the status of the flags. If the application starts offline, or if the connection is lost during runtime, feature flags fail to load, potentially causing unexpected behavior or system failures.

We request the implementation of an offline caching mechanism that will persist the last downloaded flags in the client's local storage.

keldin-coding:
Yeah, seconding that it's not great that we can't seed a React app's SDK with a snapshot on page load. We're looking at switching providers and I'd sort of assumed this was there since it was for the other SDKs I checked, but then found it's not for the JavaScript side.

markphelps:
@keldin-coding @luisbytes WIP here: https://github.com/flipt-io/flipt-client-sdks/pull/1796

markphelps:
just released. thank you for your patience

PR #1800, opened 2026-06-19
feat: add Go WASM snapshot support

markphelps:
## Summary
- add snapshot export/seed exports to the Go WASM engine
- expose Go SDK snapshot APIs via `Client.GetSnapshot` and `WithSnapshot`
- add Go snapshot tests plus JS WASM snapshot integration coverage in the Dagger JS flow

## Verification
- `cargo check -p flipt-engine-wasm`
- `cd flipt-client-go && GOWORK=off go test ./... -run 'TestUnauthorized|TestWithSnapshotInvalid|TestErrorStrategyFallback_EvaluationReturnsError'`
- `cd flipt-client-js && npm run build >/tmp/flipt-js-build-final.log 2>&1 && node integration/snapshot.mjs` (local script skipped because `FLIPT_URL` is unset; Dagger provides it)

Related to #1480.


erka (approved):
Nice,

One nit about `seed_snapshot`: should the engine check whether the snapshot has the same namespace as the engine?

PR #1802, opened 2026-06-19
feat(dart): add android armv7 packaging

### LINKED ISSUE #1801: Why are 32-bit Android architectures (armeabi-v7a, x86) not supported?
First off, thanks for building Flipt, really appreciate the work you've put into this!

I was exploring the Flutter SDK and noticed that, for Android, it only supports 64-bit architectures (`arm64-v8a` and `x86_64`). There's no support for 32-bit ABIs like `armeabi-v7a` or `x86`.

I looked through the issues and PRs but couldn't find any discussion about this, nothing in the docs or past conversations explaining why.

So I was curious:

1. **Was dropping 32-bit support a deliberate decision?** If so, what was the reasoning?
2. **Are there any plans to support 32-bit architectures in the future?**
3. **Is this a technical constraint** (e.g., the Rust FFI engine can't compile for 32-bit) or more of a **prioritization call**?

We're integrating Flipt on our Flutter app and wanted to understand the thinking here so we can act accordingly. 

Thanks again for all the hard work on this project! 🙌

markphelps:
## Summary

- add Android `armv7-linux-androideabi` / `armeabi-v7a` to FFI engine packaging
- include the new Android armv7 artifact in Dart SDK packaging under `android/src/main/jniLibs/armeabi-v7a`
- document Dart support for Android 32-bit ARM devices
- add a Go test covering Dart SDK platform support for Android armv7

Fixes #1801.

## Notes

This is intentionally limited to `armeabi-v7a` for real 32-bit ARM Android devices. It does not add Android `x86`.

Release sequencing: after this lands, a new `flipt-engine-ffi` tag needs to publish the `Android-armv7` asset before the Dart SDK package can be released against that engine tag.

## Verification

- `git diff --check HEAD~1..HEAD`
- `go test ./package/ffi/...`

I also attempted `cargo check -p flipt-engine-ffi --target armv7-linux-androideabi`; it reached Android toolchain setup and failed locally because `arm-linux-androideabi-clang` / cargo-ndk is not installed on this machine. The GitHub Android packaging workflow uses `cargo-ndk` and should be the authoritative build check for this target.


erka (approved):
nice

PR #1817, opened 2026-06-24
fix: return shared &Engine from get_engine to fix aliasing UB

yoshiyuki-nakano-safie:
## Summary

`get_engine`, the FFI entrypoint helper, returned a `&mut Engine` (an exclusive mutable reference) on every call. A single engine pointer is commonly shared across threads by FFI consumers, so evaluating the same engine concurrently created multiple aliasing `&mut Engine` to the same object — undefined behavior in Rust, which can corrupt the heap under real concurrent load.

Evaluation only needs `&self`, so this changes `get_engine` to return a shared `&Engine`.

## Problem

```rust
unsafe fn get_engine<'a>(engine_ptr: *mut c_void) -> Result<&'a mut Engine, FFIError> {
    ...
    Ok(&mut *(engine_ptr as *mut Engine))
}
```

The snapshot is correctly protected by an `Arc<RwLock<Evaluator>>`, so there is no data race on the flag state. The bug is one level up: every evaluation method (`variant`, `boolean`, `batch`, `list_flags`, `get_snapshot`) takes `&self`, yet the entrypoint materializes a `&mut Engine`. When a shared pointer is evaluated from multiple threads at once, two or more live `&mut` to the same `Engine` exist concurrently. That violates Rust's aliasing rules and is UB regardless of the lock — `rustc` applies `noalias` to `&mut`, and that assumption breaks under concurrent access. (`destroy_engine` is unaffected: it uses `Box::from_raw`, not `get_engine`.)

## Solution

Return a shared reference, since evaluation only needs `&self`.

```rust
unsafe fn get_engine<'a>(engine_ptr: *mut c_void) -> Result<&'a Engine, FFIError> {
    ...
    Ok(&*(engine_ptr as *const Engine))
}
```

Every method reachable through `get_engine` takes `&self`, so no caller needs `&mut` and the change compiles with no other modifications. The mutable state (the snapshot) stays guarded by the internal `RwLock`, so background updates keep working, and any number of threads can now evaluate the same pointer concurrently without violating aliasing rules.

## Background

I found this in a production-scale load test. A Python service shared a single client (the `ctypes` wrapper over `libfliptengine.so`) and evaluated it concurrently from many worker threads. The process aborted with glibc's `double free or corruption (!prev)` (`SIGABRT`). A core dump showed 8 threads inside the engine's evaluation path at the moment of the crash, while the thread that tripped the abort was a different one doing unrelated `malloc`/`free` — the classic signature of heap corruption surfacing on whichever thread next touches the heap.

### DIFF
--- a/flipt-engine-ffi/src/lib.rs
+++ b/flipt-engine-ffi/src/lib.rs
@@ -951,11 +951,11 @@ unsafe extern "C" fn _destroy_string(ptr: *mut c_char) {
 
 /// This function should not be called unless an Engine is initiated. It provides a helper
 /// utility to retrieve an Engine instance for evaluation use.
-unsafe fn get_engine<'a>(engine_ptr: *mut c_void) -> Result<&'a mut Engine, FFIError> {
+unsafe fn get_engine<'a>(engine_ptr: *mut c_void) -> Result<&'a Engine, FFIError> {
     if engine_ptr.is_null() {
         Err(FFIError::NullPointer)
     } else {
-        Ok(&mut *(engine_ptr as *mut Engine))
+        Ok(&*(engine_ptr as *const Engine))
     }
 }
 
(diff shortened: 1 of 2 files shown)


erka (approved):
Thank you for catching this and fixing it. @yoshiyuki-nakano-safie

markphelps:
Thank you @yoshiyuki-nakano-safie ! will create new releases of all the FFI based SDKs to include this fix

PR #1837, opened 2026-06-30
fix(python): require Python 3.11 or newer

markphelps:
## Summary

- Require Python `>=3.11,<4.0` for the Python SDK package metadata
- Update the Python SDK README migration note to say 3.11 is the minimum supported version
- Refresh the Poetry lock metadata and drop Python <3.11-only transitive entries
- Add a small test that enforces the declared Python version range in `pyproject.toml`

## Context

Python 3.10 is security-only, while Python 3.11+ remains within the supported lifecycle. This makes the SDK support policy explicit in packaging metadata instead of only relying on CI/tooling assumptions.

## Verification

- `mise exec -- poetry install`
- `mise exec -- poetry run pytest tests/test_python_version.py`
- `mise exec -- poetry run black --check .`
- Confirmed `pyproject.toml`, `poetry.lock`, and README declare Python 3.11+ and no Python `<3.11` markers remain in the Python SDK metadata/docs touched by this change.

Note: the full local Python SDK test suite still requires `FLIPT_URL`; without it, existing integration-style tests fail during setup before this change is exercised.


erka (approved):
nice one!

PR #1853, opened 2026-07-10
fix(android): package armeabi-v7a library

### LINKED ISSUE #1852: Android Client SDK missing `armeabi-v7a` native library causes `UnsatisfiedLinkError` on 32-bit ARM devices
Hi folks,

We’re seeing `UnsatisfiedLinkError` crashes on some Android devices with `io.flipt:flipt-client-android:1.3.3`.

The crash happens during `FliptClient` initialization, where `CLibrary` calls:

`System.loadLibrary("fliptengine_wrapper")`

From inspecting the published AAR, it looks like the native library is only packaged for:
- `arm64-v8a`
- `x86_64`

and not for:
- `armeabi-v7a`

That means apps can crash on unsupported 32-bit ARM devices when `FliptClient` is created.

Thanks!

markphelps:
## Summary

Fixes #1852 by including the Android 32-bit ARM (`armeabi-v7a`) native wrapper in the Kotlin Android SDK AAR.

- Include `Android-armv7` when assembling the Android SDK package and stage it under `armeabi-v7a`.
- Configure Gradle and CMake to build and link the armv7 wrapper.
- Download and stage the armv7 engine artifact in both Android SDK CI paths.
- Document the supported ABI and add a regression test for armv7 package-platform selection.

The FFI engine release pipeline already builds and publishes the `Android-armv7` artifact; this change carries it through the Kotlin Android SDK package.

## Verification

- `go test ./package/ffi/sdks`
- `cd flipt-client-kotlin-android && mise run lint`
- Parsed `.github/workflows/test-android-sdk.yml` successfully with Ruby YAML

### DIFF
--- a/.github/workflows/test-android-sdk.yml
+++ b/.github/workflows/test-android-sdk.yml
@@ -91,20 +91,32 @@ jobs:
           path: /tmp
           run-id: ${{ github.event.workflow_run.id }}
 
+      - name: Download Artifact (Android armv7)
+        uses: actions/download-artifact@v8
+        with:
+          github-token: ${{ steps.generate_token.outputs.token }}
+          name: flipt-engine-ffi-Android-armv7.tar.gz
+          path: /tmp
+          run-id: ${{ github.event.workflow_run.id }}
+
       - name: Extract Artifacts
         run: |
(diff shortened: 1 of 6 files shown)


erka (approved):
Nice. Hard to know for sure, but let's give it a try.

PR #1927, opened 2026-08-28
fix(ffi): apply default request timeout in polling mode (#1926)

### LINKED ISSUE #1926: FFI SDKs: the polling refresher stops permanently
With FetchMode.POLLING, if the engine's background refresh task stops, the client keeps evaluating the snapshot it fetched at startup.

The only recovery available to a consumer is to build a whole new FliptClient.

HTTPFetcher::start breaks out of the loop on error:

https://github.com/flipt-io/flipt-client-sdks/blob/0ebaa5025ab8c208fb15c8da9ca068187215cd6f/flipt-engine-ffi/src/http.rs#L342-L345

Nothing restarts the task, and the Engine holding it stays alive and keeps serving evaluations from the last snapshot.

**Environment**
flipt-client-java 1.3.3, FetchMode.POLLING, updateInterval 30s, default ErrorStrategy.FAIL
Flipt server v1.61.1, Postgres backend, 3 replicas

johny-cho:
## Problem
With `FetchMode.POLLING` and the default configuration, the refresher can
stop permanently and the engine keeps serving the snapshot fetched at
startup. Recovery requires building a brand new client.

## Root cause
`HTTPFetcherBuilder::build` only applies `.timeout(..)` when the caller
explicitly sets `request_timeout`. With the default (unset), the client has
`connect_timeout(10s)` but no overall request timeout. If a connection is
established but the response then stalls, `handle_polling` awaits the fetch
(and the body read) forever; the polling loop is blocked on that await and
never advances. It is a hang, not the `break` path (which only triggers when
the receiver channel is closed).

## Fix
Always apply an overall request timeout in polling mode, defaulting to 30s
when the caller did not set one. Note the builder also discards sub-second
`request_timeout` values (`as_secs() > 0`), so those fall back to the
default too rather than leaving the client without a timeout. A stalled
response now times out, surfaces as an error to the consumer, and the loop
proceeds to the next interval.

## Testing
- `test_polling_request_timeout_resolution`: unit test asserting the timeout
  resolution — default applied when unset or sub-second, explicit valid
  value respected.
- `test_polling_default_timeout_unblocks_stalled_response`: end-to-end test
  that exercises the **default** path (no explicit `request_timeout`) against
  a server that sends headers then holds the body open. Before this fix that
  request hangs indefinitely; after, it times out and errors. A test-only
  injectable default keeps it fast (~1s) instead of waiting the production
  30s.
- `cargo test`, `cargo fmt --check`, `cargo clippy --all -D warnings` pass.

Fixes #1926

### DIFF
--- a/flipt-engine-ffi/src/http.rs
+++ b/flipt-engine-ffi/src/http.rs
@@ -131,6 +131,7 @@ pub struct HTTPFetcherBuilder {
     authentication: HeaderMap,
     reference: Option<String>,
     request_timeout: Option<Duration>,
+    default_request_timeout: Duration,
     update_interval: Duration,
     mode: FetchMode,
     tls_config: Option<TlsConfig>,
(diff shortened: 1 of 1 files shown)


johnycho:
Superseded by #1928 (re-opened from the correct account). Closing this one.

johny-cho:
Superseded by #1928 (re-opened from the correct account). Closing this one.

PR #1951, opened 2026-09-11
fix(js): default updateInterval to 120s when omitted from options

bhargavprajapati949:
## Problem

The Node.js `FliptClient.init()` docs/README state that `updateInterval` defaults to 120 seconds when not provided. In practice, this only held true when `init()` was called with **no arguments at all**:

```ts
FliptClient.init(); // options === undefined → ClientOptionsFactory.default() used → updateInterval: 120
```

Default parameter values in JS/TS only apply when the argument itself is `undefined` as a whole — they don't merge missing keys into a partially provided object. So calling `init()` with any options object, even without `updateInterval`, bypassed the default entirely:

```ts
FliptClient.init({ url: 'https://flipt.example.com' });
// options.updateInterval is `undefined`
// `if (options.updateInterval && options.updateInterval > 0)` is false
// setupAutoRefresh() is never called — no polling happens at all
```

This silently breaks auto-refresh for the overwhelmingly common case of passing some options (e.g. `url`, `authentication`) without explicitly setting `updateInterval`.

## Fix

Resolve `updateInterval` from `options.updateInterval` with a `?? 120` fallback at the point of use, instead of relying on the default parameter value to supply it:

```ts
const updateInterval = options.updateInterval ?? 120;
if (updateInterval > 0) {
  client.setupAutoRefresh(updateInterval * 1_000);
}
```

This preserves the existing opt-out behavior (`updateInterval: 0` still disables auto-refresh) while making the "defaults to 120s" behavior actually match what's documented in the README, regardless of whether other options are passed.

## Testing

- `mise run lint` (via `npm run lint`) passes.
- No existing unit tests cover `FliptClient.init()`'s auto-refresh path; the change is a minimal, targeted fix to restore documented behavior.


markphelps (approved):
thank you @bhargavprajapati949 !

erka:
Thanks for your contribution @bhargavprajapati949 !

Could you please sign your commit so that the DCO check doesn’t block the merge?

PR #1965, opened 2026-09-14
fix(evaluation): return FlagDisabled for disabled boolean with no rollouts

erka:
Matches flipt-io/flipt#6515.


flipt-evaluation/src/lib.rs
```diff
@@ -373,10 +374,15 @@ pub fn boolean_evaluation(
         }
     }
 
+    let reason = match (flag.enabled, rollouts_empty) {
+        (false, true) => flipt::EvaluationReason::FlagDisabled,
```
markphelps, line 378: did not know match worked on tuples like this. neat

markphelps (approved):

